PR #7, opened 2019-07-26
Add datagram_socket class

josephnoir:
This PR adds a bare-bones datagram socket. Not sure what a datagram socket pair would look like or how to build a good test for this.

### DIFF
--- a/libcaf_net/caf/net/datagram_socket.hpp
+++ b/libcaf_net/caf/net/datagram_socket.hpp
@@ -24,7 +24,7 @@
 namespace caf {
 namespace net {
 
-/// A unconnected datagram-oriented network communication endpoint.
+/// A datagram-oriented network communication endpoint.
 struct datagram_socket : abstract_socket<datagram_socket> {
   using super = abstract_socket<datagram_socket>;
 
--- a/libcaf_net/caf/net/datagram_socket.hpp
+++ b/libcaf_net/caf/net/datagram_socket.hpp
@@ -1,0 +1,53 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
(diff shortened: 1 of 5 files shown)


libcaf_net/test/datagram_socket.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+using namespace caf;
+using namespace caf::net;
+
+CAF_TEST(invalid_socket) {
+  datagram_socket x;
+#ifdef CAF_WINDOWS
+  CAF_CHECK_EQUAL(allow_connreset(x, true), sec::network_syscall_failed);
+#else  // CAF_WINDOWS
+  CAF_CHECK_EQUAL(allow_connreset(x, true), none);
```
Neverlord: I think this should fail on non-Windows systems as well. Simply because passing `invalid_socket` should get checked (e.g. see `child_process_inherit` on Windows). Then we can check for `CAF_CHECK_NOT_EQUAL(..., none);`.

libcaf_net/caf/net/datagram_socket.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#pragma once
+
+#include "caf/net/network_socket.hpp"
+#include "caf/variant.hpp"
+
+namespace caf {
+namespace net {
+
+/// A unconnected datagram-oriented network communication endpoint.
```
Neverlord: I think `unconnected` is misleading here. If something has no notion of connections, than it also shouldn't be named unconnected IMO.

```suggestion
/// A datagram-oriented network communication endpoint.
```

Neverlord:
> Not sure what a datagram socket pair would look like or how to build a good test for this.

TBH, me neither. We don't want to add expensive indirections over UDP to make the API "generic". At the same time, adding `udp_socket` and pulling in all the low-level C types is equally unattractive. I'd say we leave `datagram_socket` as-is for now. Maybe we can generalize some parts of the UDP API later on.

Neverlord:
Oh, one more thing. I think we should also update the documentation of `socket` now: https://github.com/actor-framework/incubator/blob/master/libcaf_net/caf/net/socket.hpp#L35

PR #12, opened 2019-08-07
Add doorman policy

josephnoir:
This PR adds a doorman policy which accepts TCP connections ans hands them off to new `endpoint_manger` instances. Additionally, there is functionality to create the required sockets, to accept connections, and to connect to remote endpoints.


libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,102 @@
...
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#pragma once
+
+#include <vector>
+
+#include "caf/detail/socket_sys_includes.hpp"
```
Neverlord: This is a header we generally want to avoid to include in other headers. Is this really needed? Otherwise, can we restructure the code a bit so that this can go into a `.cpp` file instead?

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+  net::socket acceptor_;
+
+  net::socket handle() {
+    return acceptor_;
+  }
+
+  template <class Parent>
+  error init(Parent& parent) {
+    parent.application().init(parent);
```
Neverlord: We should check for an error here.

libcaf_net/caf/net/stream_socket.hpp
```diff
@@ -18,8 +18,11 @@
 
 #pragma once
 
+#include "caf/expected.hpp"
```
Neverlord: Unnecessary.

libcaf_net/caf/net/stream_socket.hpp
```diff
@@ -18,8 +18,11 @@
...
 #include "caf/fwd.hpp"
 #include "caf/net/network_socket.hpp"
+#include "caf/sec.hpp"
```
Neverlord: Unnecessary.

libcaf_net/caf/net/stream_socket.hpp
```diff
@@ -18,8 +18,11 @@
...
+#include "caf/variant.hpp"
```
Neverlord: Unnecessary.

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,102 @@
...
+namespace policy {
+
+/// A doorman accepts TCP connections and creates scribes to handle them.
+class doorman {
+public:
+  doorman(net::stream_socket acceptor) : acceptor_(acceptor) {
+    // nop
+  }
+
+  net::socket acceptor_;
```
Neverlord: Why is this a basic `socket` and not an accept socket?

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,102 @@
...
+    auto sck = net::tcp::accept(
+      net::socket_cast<net::stream_socket>(acceptor_));
+    if (!sck) {
+      CAF_LOG_ERROR("accept failed:" << sck.error());
+      return false;
+    }
+    auto mpx = parent.multiplexer();
+    if (!mpx) {
+      CAF_LOG_DEBUG(
+        "could not acquire multiplexer to create a new endpoint manager");
```
Neverlord:
```suggestion
        "unable to get multiplexer from parent");
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,102 @@
...
+      return false;
+    }
+    auto child = make_endpoint_manager(mpx, parent.system(), scribe{*sck},
+                                       parent.application().make());
+    if (auto err = child->init()) {
```
Neverlord:
```suggestion
    if (auto err = child->init())
```

libcaf_net/src/interfaces.cpp
```diff
@@ -0,0 +1,144 @@
...
+ ******************************************************************************/
+
+#include "caf/net/interfaces.hpp"
+
+#include <algorithm>
+#include <cerrno>
+#include <cstdlib>
+#include <cstring>
+
+// clang-format off
```
Neverlord: Most of this is redundant to the includes in `socket_sys_includes.hpp`. Why not add the delta there?

libcaf_net/caf/net/socket.hpp
```diff
@@ -55,6 +55,10 @@ void close(socket x);
 /// @relates socket
 std::errc last_socket_error();
 
+/// Returns the string representation of a given socket error code.
+/// @relates socket
+std::string socket_error_as_string(std::errc err);
```
Neverlord: This is never used and also lacks an implementation on Windows. Why did you add it?

libcaf_net/src/stream_socket.cpp
```diff
@@ -10,7 +10,6 @@
  * Distributed under the terms and conditions of the BSD 3-Clause License or  *
  * (at your option) under the terms and conditions of the Boost Software      *
  * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
- *                                                                            *
```
Neverlord: Why did you remove this line?

libcaf_net/caf/net/interfaces.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include "caf/optional.hpp"
+
+namespace caf {
+namespace net {
+
+/// Utility class bundling access to network interface names and addresses.
+class interfaces {
+public:
+  /// Returns a native IPv4 or IPv6 translation of `host`.
+  static optional<std::pair<std::string, ip>>
```
Neverlord: Why is this returning a `string`? The documentation states `native IPv4 or IPv6` address, so why doesn't the function return an `ip_address`?

We are designing a new API here, so please let's not simply take (bad) code from `caf_io`. 😉

libcaf_net/caf/net/interfaces.hpp
```diff
@@ -0,0 +1,44 @@
...
+  native_address(const std::string& host, optional<ip> preferred = none);
+
+  /// Returns the host and protocol available for a local server socket
+  static std::vector<std::pair<std::string, ip>>
+  server_address(uint16_t port, const char* host,
```
Neverlord: In the past, we needed this function for `publish` to work with a port only. However, we want to ditch that design anyway. Can't we drop this "guess-magic" entirely and instead encode everything we need in the config via URI? That's code I'd really like to get rid of.

libcaf_net/caf/net/ip.hpp
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include <cstddef>
+#include <string>
+
+namespace caf {
+namespace net {
+
+// IP version tag.
+enum class ip { v4, v6 };
```
Neverlord: We only need this for the `preferred` arguments. I hope we can get rid of these to also make this header obsolete.

libcaf_net/caf/net/socket_guard.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "caf/net/socket.hpp"
+
+namespace caf {
+namespace net {
+
+/// Closes the guarded socket when destroyed.
+class socket_guard {
+public:
+  explicit socket_guard(net::socket fd);
```
Neverlord: Looking at the client code, I think this class should be a template instead, because `release` is pretty useless otherwise (or forces users to fall back to `socket_cast`).

For ease of use, there also should be a factory function:

```c++
template <class Socket>
socket_guard<Socket> make_socket_guard(Socket sock) {
  return ...;
}
```

libcaf_net/caf/net/tcp.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <cstdint>
+
+#include "caf/expected.hpp"
+#include "caf/net/ip.hpp"
+#include "caf/net/stream_socket.hpp"
+
+namespace caf {
+namespace net {
+
+struct tcp {
```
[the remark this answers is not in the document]
josephnoir: I can get behind the `tcp_accept_socket`, but what does the `tcp_stream_socket` offer us? The read and write functions won't work on it unless you convert it to a `stream_socket`. And what is the use case for the `stream_socket` if we have specific types for TCP?

libcaf_net/caf/policy/doorman.hpp
josephnoir: Just forgot to remove this. It was necessary before moving `accept` into a separate function.

libcaf_net/caf/policy/doorman.hpp
josephnoir: Copied this from the scribe. I'll add a check there as well.

libcaf_net/caf/net/stream_socket.hpp
josephnoir: There is a function here that has an expected type as a return value. Is the implicit include good?

libcaf_net/src/stream_socket.cpp
josephnoir: Really good question.

libcaf_net/caf/net/stream_socket.hpp
Neverlord: Declaring a function prototype always only requires forward declarations for return type and argument types.

libcaf_net/caf/net/ip.hpp
josephnoir: The argument currently has three states. Prefer IPv4, prefer IPv6, and don't prefer anything. Since it already is optional we could just as well make it an optional bool named `prefer_v6` that is set to `none`. Or are you suggesting to change the functionality here?

josephnoir: Probably solves itself if we find a better way to address the comment above.

libcaf_net/caf/net/socket_guard.hpp
```diff
@@ -0,0 +1,62 @@
...
+    close();
+  }
+
+  Socket release() {
+    auto fd = fd_;
+    fd_ = Socket{};
+    return fd;
+  }
+
+  void close() {
```
Neverlord: Is there any use case for calling this on the guard? I think the code can safely go into the destructor.

libcaf_net/caf/net/socket_manager.hpp
```diff
@@ -52,6 +52,11 @@ class socket_manager : public ref_counted {
     return handle_;
   }
 
+  /// Returns a pointer to the multiplexer running the socket manager.
```
Neverlord: This upgrades a weak pointer to a strong pointer. So the documentation should at least mention that the function can `nullptr` and why.

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "caf/net/network_socket.hpp"
+#include "caf/net/socket.hpp"
+#include "caf/net/stream_socket.hpp"
+#include "caf/net/tcp_stream_socket.hpp"
+#include "caf/uri.hpp"
+
+namespace caf {
+namespace net {
+
+/// Represents an open TCP endpoint. Can be implicitly converted to a
```
Neverlord: I would drop the remarks on implicit conversions. Also, I find "open TCP endpoint" a bit confusion. Why not simply this?

```suggestion
/// Represents a TCP acceptor in listening mode.
```

josephnoir:
@Neverlord Would you mind giving this another look?

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  constexpr operator socket() const noexcept {
+    return socket{id};
+  }
+
+  constexpr operator network_socket() const noexcept {
+    return network_socket{id};
+  }
+
+  constexpr operator stream_socket() const noexcept {
```
Neverlord: Acceptors cannot send or receive data.

```suggestion

```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+    return stream_socket{id};
+  }
+};
+
+/// Creates a new TCP socket to accept connections on a given port.
+/// @param port The port to listen on.
+/// @param addr Only accepts connections originating from this address.
```
Neverlord:
```suggestion
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+/// @param reuse_addr Optionally sets the SO_REUSEADDR option on the socket.
+/// @relates stream_socket
+expected<tcp_accept_socket> make_accept_socket(ip_address host, uint16_t port,
+                                               bool reuse_addr);
```
Neverlord:
```suggestion
                                               bool reuse_addr = false);
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+/// Creates a new TCP socket to accept connections on a given port.
+/// @param port The port to listen on.
```
Neverlord:
```suggestion
/// @param auth ...
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+/// @param addr Only accepts connections originating from this address.
```
Neverlord:
```suggestion
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+/// @param reuse_addr Optionally sets the SO_REUSEADDR option on the socket.
```
Neverlord:
```suggestion
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+/// @relates stream_socket
```
Neverlord:
```suggestion
/// @relates tcp_accept_socket
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+  constexpr operator stream_socket() const noexcept {
+    return stream_socket{id};
+  }
+};
+
+/// Creates a new TCP socket to accept connections on a given port.
+/// @param port The port to listen on.
+/// @param addr Only accepts connections originating from this address.
+/// @param reuse_addr Optionally sets the SO_REUSEADDR option on the socket.
+/// @relates stream_socket
```
Neverlord:
```suggestion
/// @relates tcp_accept_socket
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+expected<tcp_accept_socket> make_accept_socket(const uri::authority_type& auth,
+                                               bool reuse_addr = false);
+
+/// Accept a connection on `x`.
```
Neverlord:
```suggestion
/// Tries to accept a connection on `x`.
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+/// @param x Listening endpoint.
+/// @returns The socket that handles the accepted connection.
```
Neverlord:
```suggestion
/// @returns The socket that handles the accepted connection on success, an error otherwise.
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+/// @relates stream_socket
```
Neverlord:
```suggestion
/// @relates tcp_stream_socket
/// @relates tcp_accept_socket
```

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "caf/net/abstract_socket.hpp"
+#include "caf/net/network_socket.hpp"
+#include "caf/net/socket.hpp"
+#include "caf/net/stream_socket.hpp"
+#include "caf/uri.hpp"
+
+namespace caf {
+namespace net {
+
+/// Represents a TCP connection. Can be implicitly converted to a
```
Neverlord: Here, I'd also drop the remarks on implicit conversion.

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+  }
+};
+
+/// Create a `stream_socket` connected to `host`:`port`.
+/// @param host The remote host to connecto to.
+/// @param port The port on the remote host to connect to.
+/// @preferred Preferred IP version.
+/// @returns The connected socket or an error.
+/// @relates stream_socket
+expected<stream_socket> make_connected_socket(ip_address host, uint16_t port);
```
Neverlord:
```suggestion
expected<tcp_stream_socket> make_connected_tcp_stream_socket(ip_endpoint node);
```

It's quite a mouthful, but at least it's accurate.

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+  constexpr operator stream_socket() const noexcept {
+    return stream_socket{id};
+  }
+};
+
+/// Create a `stream_socket` connected to `host`:`port`.
+/// @param host The remote host to connecto to.
+/// @param port The port on the remote host to connect to.
+/// @preferred Preferred IP version.
```
Neverlord:
```suggestion
```

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// @returns The connected socket or an error.
+/// @relates stream_socket
```
Neverlord:
```suggestion
/// @relates tcp_stream_socket
```

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+  constexpr operator network_socket() const noexcept {
+    return network_socket{id};
+  }
+
+  constexpr operator stream_socket() const noexcept {
+    return stream_socket{id};
+  }
+};
+
+/// Create a `stream_socket` connected to `host`:`port`.
```
Neverlord:
```suggestion
/// Creates a `tcp_stream_socket` connected to given remote node.
```

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// @param host The remote host to connecto to.
+/// @param port The port on the remote host to connect to.
+/// @preferred Preferred IP version.
+/// @returns The connected socket or an error.
+/// @relates stream_socket
+expected<stream_socket> make_connected_socket(ip_address host, uint16_t port);
+
+/// Create a `stream_socket` connected to `auth`.
```
Neverlord:
```suggestion
/// Create a `tcp_stream_socket` connected to `auth`.
```

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// @param authority The remote host to connecto to.
```
Neverlord:
```suggestion
/// @param node Host and port of the remote node.
```

libcaf_net/caf/net/tcp_stream_socket.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// @preferred Preferred IP version.
+/// @returns The connected socket or an error.
+/// @relates stream_socket
+expected<stream_socket> make_connected_socket(const uri::authority_type& auth);
```
Neverlord:
```suggestion
expected<stream_socket> make_connected_socket(const uri::authority_type& node);
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+    return stream_socket{id};
+  }
+};
+
+/// Creates a new TCP socket to accept connections on a given port.
+/// @param port The port to listen on.
+/// @param addr Only accepts connections originating from this address.
+/// @param reuse_addr Optionally sets the SO_REUSEADDR option on the socket.
+/// @relates stream_socket
+expected<tcp_accept_socket> make_accept_socket(ip_address host, uint16_t port,
```
Neverlord:
```suggestion
expected<tcp_accept_socket> make_accept_socket(ip_endpoint node,
```

libcaf_net/caf/net/tcp_accept_socket.hpp
```diff
@@ -0,0 +1,75 @@
...
+    return network_socket{id};
+  }
+
+  constexpr operator stream_socket() const noexcept {
+    return stream_socket{id};
+  }
+};
+
+/// Creates a new TCP socket to accept connections on a given port.
+/// @param port The port to listen on.
```
Neverlord:
```suggestion
/// @param node The port to listen on and the filter for incoming addresses. Passing the address `0.0.0.0` will accept incoming connection from any host.
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,100 @@
...
+/// A doorman accepts TCP connections and creates scribes to handle them.
+class doorman {
+public:
+  doorman(net::tcp_accept_socket acceptor) : acceptor_(acceptor) {
+    // nop
+  }
+
+  net::tcp_accept_socket acceptor_;
+
+  net::socket handle() {
```
Neverlord:
```suggestion
  net::tcp_accept_socket handle() {
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,100 @@
...
+    return acceptor_;
+  }
+
+  template <class Parent>
+  error init(Parent& parent) {
+    if (auto err = parent.application().init(parent))
+      return std::move(err);
```
Neverlord: Returning a local variable is always a `move`.

```suggestion
      return err;
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,100 @@
...
+    parent.mask_add(net::operation::read);
+    return none;
+  }
+
+  template <class Parent>
+  bool handle_read_event(Parent& parent) {
+    auto sck = net::accept(acceptor_);
+    if (!sck) {
+      CAF_LOG_ERROR("accept failed:" << sck.error());
```
Neverlord:
```suggestion
      CAF_LOG_ERROR("accept failed:" << parent.system().render(sock.error()));
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,100 @@
...
+  error init(Parent& parent) {
+    if (auto err = parent.application().init(parent))
+      return std::move(err);
+    parent.mask_add(net::operation::read);
+    return none;
+  }
+
+  template <class Parent>
+  bool handle_read_event(Parent& parent) {
+    auto sck = net::accept(acceptor_);
```
Neverlord: I think we should not introduce new abbreviations:

```sh
fgrep -R 'sck' libcaf_net | wc -l
       0
```

```suggestion
    auto sock = net::accept(acceptor_);
```

Also ok IMO:

```c++
    auto x = net::accept(acceptor_);
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,100 @@
...
+  template <class Parent>
+  bool handle_write_event(Parent&) {
+    CAF_LOG_ERROR("doorman received write event");
+    return false;
+  }
+
+  template <class Parent>
+  void resolve(Parent&, const std::string& path, actor listener) {
+    CAF_LOG_ERROR("doorman called to resolve" << CAF_ARG(path));
+    anon_send(listener, resolve_atom::value, "doorman cannot resolve paths");
```
Neverlord:
```suggestion
    anon_send(listener, resolve_atom::value, "doormen cannot resolve paths");
```

libcaf_net/caf/policy/doorman.hpp
```diff
@@ -0,0 +1,100 @@
...
+  template <class Parent>
+  void timeout(Parent&, atom_value x, uint64_t id) {
+    CAF_LOG_ERROR("doorman received timeout" << CAF_ARG(x) << CAF_ARG(id));
+    CAF_IGNORE_UNUSED(x);
+    CAF_IGNORE_UNUSED(id);
+  }
+
+  template <class Application>
+  void handle_error(Application&, sec) {
+    close(acceptor_);
```
Neverlord: We already do this in the destructor. Logging the error code would be nice.

```suggestion
```

libcaf_net/caf/policy/scribe.hpp
```diff
@@ -43,7 +43,8 @@ class scribe {
 
   template <class Parent>
   error init(Parent& parent) {
-    parent.application().init(parent);
+    if (auto err = parent.application().init(parent))
+      return std::move(err);
```
Neverlord:
```suggestion
      return err;
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "caf/logger.hpp"
+#include "caf/string_algorithms.hpp"
+#include "caf/string_view.hpp"
+
+// clang-format off
+#ifdef CAF_WINDOWS
+#  ifndef _WIN32_WINNT
+#    define _WIN32_WINNT 0x0600
+#  endif
+#  include <iostream>
```
Neverlord: Why do we need `<iostream>` on Windows?

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+using std::vector;
+
+namespace caf {
+namespace net {
+namespace ip {
+
+namespace {
+
+template <class T>
+void* vptr(T* ptr) {
```
Neverlord: This function is unnecessary. Casts from any pointer to `void*` are implicit in C++.

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+  return static_cast<void*>(ptr);
+}
+
+void* fetch_in_addr(int family, sockaddr* addr) {
+  if (family == AF_INET)
+    return vptr(&reinterpret_cast<sockaddr_in*>(addr)->sin_addr);
```
Neverlord:
```suggestion
    return &reinterpret_cast<sockaddr_in*>(addr)->sin_addr;
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+  return vptr(&reinterpret_cast<sockaddr_in6*>(addr)->sin6_addr);
```
Neverlord:
```suggestion
  return &reinterpret_cast<sockaddr_in6*>(addr)->sin6_addr;
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return {};
+  std::unique_ptr<addrinfo, decltype(freeaddrinfo)*> addrs{tmp, freeaddrinfo};
+  char buffer[INET6_ADDRSTRLEN];
+  std::vector<ip_address> results;
+  for (auto i = addrs.get(); i != nullptr; i = i->ai_next) {
+    auto family = fetch_addr_str(true, true, buffer, i->ai_addr);
+    if (family != AF_UNSPEC) {
+      ip_address ip;
+      if (auto err = parse(buffer, ip)) {
+        CAF_LOG_ERROR("could not parse into ip address " << buffer);
```
Neverlord:
```suggestion
        CAF_LOG_ERROR("could not parse IP address:" << buffer);
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+        continue;
```
Neverlord:
```suggestion
        else results.emplace_back(ip);;
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+  return results;
+}
+
+#ifdef CAF_WINDOWS
+
+std::string hostname() {
+  TCHAR buf[MAX_COMPUTERNAME_LENGTH + 1];
+  DWORD size = MAX_COMPUTERNAME_LENGTH;
+  GetComputerName(buf, &size);
+  return std::string{buf};
```
Neverlord:
```suggestion
  return buf;
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#else // CAF_WINDOWS
+
+std::string hostname() {
+  char buf[HOST_NAME_MAX + 1];
+  buf[HOST_NAME_MAX] = '\0';
+  gethostname(buf, HOST_NAME_MAX);
+  gethostbyname(buf);
+  return std::string{buf};
```
Neverlord:
```suggestion
  return buf;
```

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include "caf/net/socket_guard.hpp"
+#include "caf/net/tcp_stream_socket.hpp"
+#include "caf/sec.hpp"
+
+namespace caf {
+namespace net {
+
+namespace {
+
+expected<void> set_inaddr_any(socket, sockaddr_in& sa) {
```
Neverlord:
```suggestion
error set_inaddr_any(socket, sockaddr_in& sa) {
```

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -0,0 +1,145 @@
...
+  sa.sin_addr.s_addr = INADDR_ANY;
+  return unit;
+}
+
+expected<void> set_inaddr_any(socket x, sockaddr_in6& sa) {
```
Neverlord:
```suggestion
error set_inaddr_any(socket x, sockaddr_in6& sa) {
```

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -0,0 +1,145 @@
...
+                               reinterpret_cast<setsockopt_ptr>(&on),
+                               static_cast<socket_size_type>(sizeof(on))));
+  }
+  using sockaddr_type = typename std::conditional<
+    Family == AF_INET, sockaddr_in, sockaddr_in6>::type;
+  sockaddr_type sa;
+  memset(&sa, 0, sizeof(sockaddr_type));
+  detail::family_of(sa) = Family;
+  if (any)
+    set_inaddr_any(fd, sa);
```
Neverlord:
```suggestion
    if (auto err = set_inaddr_any(fd, sa))return err;
```

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -0,0 +1,145 @@
...
+namespace {
+
+expected<void> set_inaddr_any(socket, sockaddr_in& sa) {
+  sa.sin_addr.s_addr = INADDR_ANY;
+  return unit;
+}
+
+expected<void> set_inaddr_any(socket x, sockaddr_in6& sa) {
+  sa.sin6_addr = in6addr_any;
+  // also accept ipv4 requests on this socket
```
Neverlord:
```suggestion
  // Also accept IPv4 connections on this socket.
```

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -0,0 +1,145 @@
...
+  int off = 0;
+  CAF_NET_SYSCALL("setsockopt", res, !=, 0,
+                  setsockopt(x.id, IPPROTO_IPV6, IPV6_V6ONLY,
+                             reinterpret_cast<setsockopt_ptr>(&off),
+                             static_cast<socket_size_type>(sizeof(off))));
+  return unit;
+}
+
+template <int Family, int SockType = SOCK_STREAM>
+caf::expected<socket> new_ip_acceptor_impl(uint16_t port, const char* addr,
```
Neverlord:
```suggestion
caf::expected<tcp_accept_socket> new_tcp_acceptor_impl(uint16_t port, const char* addr,
```

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -0,0 +1,145 @@
...
+                       static_cast<socket_size_type>(sizeof(sa))));
+  return sguard.release();
+}
+
+} // namespace
+
+expected<tcp_accept_socket> make_accept_socket(ip_address addr, uint16_t port,
+                                               bool reuse_addr) {
+  CAF_LOG_TRACE(CAF_ARG(port) << ", addr = " << to_string(addr));
+  auto h = addr.embeds_v4() ? to_string(addr.embedded_v4()) : to_string(addr);
```
Neverlord: It's unnecessary to dispatch `to_string`. That's already taken care of internally.

```suggestion
  auto host = to_string(addr);
```

josephnoir:
Thank you for taking the time. I'll clean it up and open smaller PRs.
*Note to self: Pay more attention.*

libcaf_net/test/doorman.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+namespace {
+
+struct fixture : test_coordinator_fixture<>, host_fixture {
+  fixture() {
+    mpx = std::make_shared<multiplexer>();
+    if (auto err = mpx->init())
+      CAF_FAIL("mpx->init failed: " << sys.render(err));
+    auth.port = 0;
+    auth.host = std::string{"0.0.0.0"};
```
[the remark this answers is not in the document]
josephnoir: The host is a `variant<string, ip_address>` and the compiler complains if I don't wrap it in a string first. How do I do this?

PR #14, opened 2019-08-19
Fix mixin for string_application

jakobod:
This PR fixes the templates for Mixins in the string_application test.

### DIFF
--- a/libcaf_net/test/string_application.cpp
+++ b/libcaf_net/test/string_application.cpp
@@ -188,6 +188,8 @@ private:
 CAF_TEST_FIXTURE_SCOPE(endpoint_manager_tests, fixture)
 
 CAF_TEST(receive) {
+  using application_type = extend<string_application>::with<
+    stream_string_application>;
   std::vector<char> read_buf(1024);
   CAF_CHECK_EQUAL(mpx->num_socket_managers(), 1u);
   auto buf = std::make_shared<std::vector<char>>();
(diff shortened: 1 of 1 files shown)


josephnoir:
Since I'm not entirely sure what the implication is when not using the second parameter for extend (does this result in the type the same type for `Subtype` and `Base`?), would you mind taking a look at this @Neverlord?

libcaf_net/test/string_application.cpp
```diff
@@ -194,12 +197,14 @@ CAF_TEST(receive) {
                   sec::unavailable_or_would_block);
   CAF_MESSAGE("adding both endpoint managers");
   auto mgr1 = make_endpoint_manager(mpx, sys, policy::scribe{sockets.first},
-                                    stream_string_application{sys, buf});
+                                    extend<string_application>::with<
```
Neverlord: This rather unwieldy type is used twice. I'd give it a type alias to make the code more readable and less repetitive:

```c++
  using application_type = extend<string_application>::with<stream_string_application>;
```

Neverlord:
> would you mind taking a look at this @Neverlord?

Not at all. 🙂

> Since I'm not entirely sure what the implication is when not using the second parameter for extend (does this result in the type the same type for Subtype and Base?)

Some mixins use the second argument for CRTP. In this case, not setting the optional second argument is fine.

libcaf_net/test/string_application.cpp
jakobod: Thanks for the tip @Neverlord.

PR #15, opened 2019-08-28
Add function to resolve hosts to IP address

josephnoir:


libcaf_net/caf/net/ip.hpp
```diff
@@ -0,0 +1,38 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright 2011-2018 Dominik Charousset                                     *
```
Neverlord:
```suggestion
 * Copyright 2011-2019 Dominik Charousset                                     *
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,136 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright 2011-2018 Dominik Charousset                                     *
```
Neverlord:
```suggestion
 * Copyright 2011-2019 Dominik Charousset                                     *
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,136 @@
...
+namespace {
+
+void* fetch_in_addr(int family, sockaddr* addr) {
+  if (family == AF_INET)
+    return &reinterpret_cast<sockaddr_in*>(addr)->sin_addr;
+  return &reinterpret_cast<sockaddr_in6*>(addr)->sin6_addr;
+}
+
+// TODO: Use getnameinfo instead?
+int fetch_addr_str(bool get_ipv4, bool get_ipv6, char (&buf)[INET6_ADDRSTRLEN],
```
Neverlord: The flags are unnecessary: The only function that ever calls `fetch_addr_str` always sets both to `true`.

```suggestion
int fetch_addr_str(char (&buf)[INET6_ADDRSTRLEN],
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,136 @@
...
+                   sockaddr* addr) {
+  if (addr == nullptr)
+    return AF_UNSPEC;
+  auto family = addr->sa_family;
+  auto in_addr = fetch_in_addr(family, addr);
+  return ((family == AF_INET && get_ipv4) || (family == AF_INET6 && get_ipv6))
+             && inet_ntop(family, in_addr, buf, INET6_ADDRSTRLEN) == buf
```
Neverlord: We should include `<arpa/inet.h>` before using `inet_ntop` [on POSIX](http://pubs.opengroup.org/onlinepubs/009695399/functions/inet_ntop.html).

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// TODO: Use getnameinfo instead?
+int fetch_addr_str(bool get_ipv4, bool get_ipv6, char (&buf)[INET6_ADDRSTRLEN],
+                   sockaddr* addr) {
+  if (addr == nullptr)
+    return AF_UNSPEC;
+  auto family = addr->sa_family;
+  auto in_addr = fetch_in_addr(family, addr);
+  return ((family == AF_INET && get_ipv4) || (family == AF_INET6 && get_ipv6))
```
Neverlord:
```suggestion
  return (family == AF_INET || family == AF_INET6)
```

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,136 @@
...
+  return buf;
+}
+
+#else // CAF_WINDOWS
+
+std::string hostname() {
+  char buf[HOST_NAME_MAX + 1];
+  buf[HOST_NAME_MAX] = '\0';
+  gethostname(buf, HOST_NAME_MAX);
+  gethostbyname(buf);
```
Neverlord: What is the purpose of calling `gethostbyname`?

libcaf_net/src/ip.cpp
```diff
@@ -0,0 +1,136 @@
...
+  //                  });
+  return results;
+}
+
+#ifdef CAF_WINDOWS
+
+std::string hostname() {
+  TCHAR buf[MAX_COMPUTERNAME_LENGTH + 1];
+  DWORD size = MAX_COMPUTERNAME_LENGTH;
+  GetComputerName(buf, &size);
```
Neverlord: The winsock API supports `gethostname`: https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-gethostname. Why aren't we using that instead of working with `ifdef`?

libcaf_net/test/ip.cpp
```diff
@@ -0,0 +1,39 @@
...
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#define CAF_SUITE ip
+
+#include "caf/net/ip.hpp"
+
+#include "caf/ipv4_address.hpp"
+#include "caf/ip_address.hpp"
+#include "caf/test/dsl.hpp"
```
Neverlord: In our other tests, we include the testing stuff before STL/CAF headers:

```c++
#include "caf/net/ip.hpp"

#include "caf/test/dsl.hpp"
	
#include "caf/ipv4_address.hpp"
#include "caf/ip_address.hpp"
```

libcaf_net/test/ip.cpp
```diff
@@ -0,0 +1,39 @@
...
+using namespace caf;
+using namespace caf::net;
+
+CAF_TEST(resolve) {
+  ip_address v4_local{make_ipv4_address(127, 0, 0, 1)};
+  ip_address v6_local{{0}, {0x1}};
+  auto addrs = ip::resolve("localhost");
+  CAF_CHECK(!addrs.empty());
+  auto contains = [&](ip_address x) {
+    return std::find(std::begin(addrs), std::end(addrs), x) != std::end(addrs);
```
Neverlord: I think using `count` here instead makes the code a bit easier to read. Also, I'm all for free functions, but the `std::` qualifiers add too much noise to be worth it IMO.

```suggestion
    return std::count(addrs.begin(), addrs.end(), x) > 0;
```

libcaf_net/src/ip.cpp
josephnoir: I want to say I tried this and got a deprecated warning.

Neverlord: `gethostbyname` is deprecated, but that's probably unnecessary anyway (see other comment).

mavam:
That's a useful feature and something we plan on using in the medium term. Have you considered an asynchronous API (with actors) to resolve addresses? Blocking on `gethostbyname` is quite expensive.

Neverlord:
@mavam we can always wrap expensive calls in actors:

```c++
  auto resolver = [](event_based_actor* self, actor parent, std::string host) {
    self->send(parent, resolve(host));
  };
  // ... in some actor ...
  self->system().spawn(resolver, self, "my.domain");
```

Or do you have something more advanced in mind?

mavam:
> Or do you have something more advanced in mind?

What I actually meant is decouple the request and the response of the name resolution process, such as in [libuv](http://docs.libuv.org/en/v1.x/dns.html). Exposing a DNS actor is a sensible abstraction, but then the implementation internally should eventually also operate truly asynchronously.

PR #16, opened 2019-08-28
Add socket guard

josephnoir:


libcaf_net/caf/net/socket_guard.hpp
```diff
@@ -0,0 +1,58 @@
...
+/// Closes the guarded socket when destroyed.
+template <class Socket>
+class socket_guard {
+public:
+  explicit socket_guard(Socket sock) : sock_(sock) {
+    // nop
+  }
+
+  ~socket_guard() {
+    using net::close;
```
Neverlord: I know I told to put it in. But it's actually not necessary. 😉

I didn't consider that we're already in namespace `caf::net`. So the compiler will look there anyway.

```suggestion

```

libcaf_net/caf/net/socket_guard.hpp
```diff
@@ -0,0 +1,58 @@
...
+    if (sock_.id != invalid_socket_id) {
+      close(sock_);
+      sock_.id = invalid_socket_id;
```
Neverlord: Unnecessary write. No one's ever going to read that value (except the unit test where `id` is a reference).

```suggestion

```

libcaf_net/test/socket_guard.cpp
```diff
@@ -0,0 +1,81 @@
...
+  {
+    auto guard = make_socket_guard(sock);
+    CAF_CHECK_EQUAL(sock.id, dummy_id);
+  }
+  CAF_CHECK_EQUAL(sock.id, invalid_socket_id);
+  CAF_CHECK(sock.closed);
+}
+
+CAF_TEST(release) {
+  auto id = dummy_id;
```
Neverlord: Don't repeat yourself: `id`, `closed`, and `sock` should be part of a fixture.

libcaf_net/caf/net/socket_guard.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#pragma once
+
+#include "caf/net/socket.hpp"
```
Neverlord: All we need is the constant `invalid_socket_id`.

```suggestion
#include "caf/net/socket_id.hpp"
```

PR #18, opened 2019-08-29
 Refactor scribe to stream-transport

jakobod:
This PR refactors the previously named `scribe` to `stream_transport`. This goes hand in hand with some design changes regarding `endpoint_manager` and the new `transport_worker` class.


libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -25,26 +25,48 @@
...
     return handle_;
   }
 
+  // -- member functions -------------------------------------------------------
+
   template <class Parent>
   error init(Parent& parent) {
-    parent.application().init(parent);
-    parent.mask_add(net::operation::read);
+    worker_.init(parent);
```
Neverlord:
```suggestion
    if (auto err = worker_.init(parent))
      return err;
```

libcaf_net/caf/net/transport_worker.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include "caf/net/endpoint_manager.hpp"
+#include "caf/net/fwd.hpp"
+#include "caf/net/write_packet_decorator.hpp"
+#include "caf/span.hpp"
+#include "caf/unit.hpp"
+
+namespace caf {
+namespace net {
+
+/// implements a worker for the udp_transport policy
```
Neverlord:
```suggestion
/// Implements a worker for transport protocols.
```

libcaf_net/caf/net/transport_worker.hpp
```diff
@@ -0,0 +1,98 @@
...
+template <class Application, class IdType = unit_t>
+class transport_worker {
```
Neverlord: This class lacks a forward declaration in `fwd.hpp`.

libcaf_net/caf/net/write_packet_decorator.hpp
```diff
@@ -26,34 +29,44 @@ namespace net {
...
+  transport_type& transport() {
+    return parent_.transport();
   }
 
-  void set_timeout(timestamp tout, atom_value type, uint64_t id) {
-    parent_.set_timeout(tout, type, id);
+  // -- member functions -------------------------------------------------------
+
+  template <class Header, class... Ts>
+  void write_packet(const Header& header, span<const byte> payload,
```
Neverlord: We always pass the header as byte span:

```suggestion
  void write_packet(span<const byte> header, span<const byte> payload,
```

libcaf_net/test/string_application.cpp
```diff
@@ -100,12 +100,11 @@ class string_application {
...
-    std::vector<byte> buf;
     header_type header{static_cast<uint32_t>(msg->payload.size())};
-    buf.resize(sizeof(header_type));
-    memcpy(buf.data(), &header, buf.size());
-    buf.insert(buf.end(), msg->payload.begin(), msg->payload.end());
-    parent.write_packet(buf);
+    std::vector<byte> header_buf(sizeof(header_type));
+    std::vector<byte> payload(msg->payload.begin(), msg->payload.end());
+    memcpy(header_buf.data(), &header, header_buf.size());
+    parent.write_packet(make_span(header_buf), make_span(payload));
```
Neverlord: I think `header_buf` is unnecessary:

```suggestion
    parent.write_packet(as_bytes(make_span(&header, 1)), make_span(payload));
```

libcaf_net/test/transport_worker.cpp
```diff
@@ -0,0 +1,222 @@
...
+  using transport_type = dummy_transport;
+
+  using application_type = dummy_application;
+
+  dummy_transport(std::shared_ptr<transport_result> res) : res_(res) {
+    // nop
+  }
+
+  template <class Header>
+  void write_packet(Header header, span<const byte> payload, ip_endpoint ep) {
```
Neverlord: We know the type of `Header`, no need for a template. 🙂

```suggestion
  void write_packet(span<const byte> header, span<const byte> payload, ip_endpoint ep) {
```

libcaf_net/test/transport_worker.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+CAF_TEST_FIXTURE_SCOPE(endpoint_manager_tests, fixture)
+
+CAF_TEST(construction and initialization) {
+  CAF_CHECK_EQUAL(worker.init(transport), none);
+  CAF_CHECK_EQUAL(application_results->initialized, true);
+}
+
+CAF_TEST(handle_data) {
+  auto test_span = make_span(reinterpret_cast<byte*>(
```
Neverlord:
```suggestion
  auto test_span = as_bytes(make_span(hello_test));
```

jakobod: This does not work. `handle_data()` requires a `span<byte>` not a `span<const byte>`. This wont compile with this fix.
Error:
```
invalid conversion from ‘caf::span<const caf::byte>::pointer’ {aka ‘const caf::byte*’} to ‘caf::span<caf::byte>::pointer’ {aka ‘caf::byte*’} [-fpermissive]
```

Neverlord: Take a look at the API: there's `as_writable_bytes` for that.

jakobod: I missed that, but still.. I cant cast away `const` using a `reinterpret_cast` I have to use `const_cast<byte*>(ptr)` at some point, or am I missing something else here?

Neverlord: Isn't the error that `transport_worker` takes `span<byte>` as argument to `handle_data` instead of `span<const byte>`? To me it looks like giving a mutable buffer to the worker shouldn't happen here, or am I missing something?

jakobod: That was it. I fixed it. Thank you! :+1:

libcaf_net/caf/net/fwd.hpp
```diff
@@ -27,6 +27,8 @@ namespace net {
 
 class multiplexer;
 class socket_manager;
+template <class Application, class IdType>
```
Neverlord:
```suggestion
template <class Application, class IdType = unit_t>
```

then don't forget to include `fwd.hpp` in `transport_worker.hpp` (the implementation of the template then does *not* declare the default again).

libcaf_net/caf/net/transport_worker.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include "caf/net/fwd.hpp"
+#include "caf/net/write_packet_decorator.hpp"
+#include "caf/span.hpp"
+#include "caf/unit.hpp"
+
+namespace caf {
+namespace net {
+
+/// Implements a worker for transport protocols.
+template <class Application, class IdType = unit_t>
```
Neverlord:
```suggestion
template <class Application, class IdType>
```

PR #19, opened 2019-08-30
Add udp_datagram_socket

jakobod:
This PR adds a `udp_datagram_socket`.


libcaf_net/src/convert_ip_endpoint.cpp
```diff
@@ -0,0 +1,60 @@
...
+  return sockaddr;
+}
+
+ip_endpoint to_ip_endpoint(const sockaddr_storage& addr) {
+  if (addr.ss_family == AF_INET) {
+    auto sockaddr4 = reinterpret_cast<const sockaddr_in*>(&addr);
+    ipv4_address ipv4_addr;
+    memcpy(ipv4_addr.data().data(), &sockaddr4->sin_addr, ipv4_addr.size());
+    return {ipv4_addr, htons(sockaddr4->sin_port)};
+  } else {
```
Neverlord:
```suggestion
  } else if (addr.ss_family == AF_INET6) {
```

libcaf_net/src/convert_ip_endpoint.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto sockaddr6 = reinterpret_cast<const sockaddr_in6*>(&addr);
+    ipv6_address ipv6_addr;
+    memcpy(ipv6_addr.bytes().data(), &sockaddr6->sin6_addr,
+           ipv6_addr.bytes().size());
+    return {ipv6_addr, htons(sockaddr6->sin6_port)};
+  }
```
Neverlord:
```suggestion
  }
  // ... error handling ...
```

libcaf_net/src/udp_datagram_socket.cpp
```diff
@@ -0,0 +1,137 @@
...
+                                                  span<byte> buf) {
+  sockaddr_storage addr = {};
+  socklen_t len = sizeof(sockaddr_storage);
+  auto res = ::recvfrom(x.id, reinterpret_cast<socket_recv_ptr>(buf.data()),
+                        buf.size(), 0, reinterpret_cast<sockaddr*>(&addr),
+                        &len);
+  auto ret = check_udp_datagram_socket_io_res(res);
+  if (auto num_bytes = get_if<size_t>(&ret)) {
+    if (*num_bytes == 0)
+      CAF_LOG_INFO("Received empty datagram");
```
Neverlord: The if-else tree only affects logging. I'd use the proper logging macros here instead:

```suggestion
      CAF_LOG_INFO_IF(*num_bytes == 0, "Received empty datagram");
```

libcaf_net/src/udp_datagram_socket.cpp
```diff
@@ -0,0 +1,137 @@
...
+    else if (*num_bytes > buf.size())
+      CAF_LOG_WARNING("recvfrom cut of message, only received "
```
Neverlord:
```suggestion
      CAF_LOG_WARNING_IF(*num_bytes > buf.size(), "recvfrom cut of message, only received "
```

libcaf_net/test/convert_ip_endpoint.cpp
```diff
@@ -0,0 +1,70 @@
...
+  ipv6_endpoint source_ep;
+  if (auto err = detail::parse("[::1]:55555", source_ep))
+    CAF_FAIL("unable to parse input: " << err);
+  auto addr = to_sockaddr(source_ep);
+  auto dest_ep = to_ip_endpoint(addr);
+  CAF_CHECK_EQUAL(source_ep, dest_ep);
+}
+
+CAF_TEST(sockaddr_in4 roundtrip) {
+  sockaddr_in source_addr = {};
```
Neverlord:
```suggestion
  sockaddr_in source_addr;
```

libcaf_net/test/convert_ip_endpoint.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "caf/detail/socket_sys_includes.hpp"
+#include "caf/ipv4_endpoint.hpp"
+#include "caf/ipv6_endpoint.hpp"
+
+using namespace caf;
+using namespace caf::detail;
+
+CAF_TEST(sockaddr_in6 roundtrip) {
+  sockaddr_in6 source_addr = {};
```
Neverlord:
```suggestion
  sockaddr_in6 source_addr;
```

libcaf_net/test/convert_ip_endpoint.cpp
```diff
@@ -0,0 +1,70 @@
...
+  source_addr.sin6_family = AF_INET6;
+  source_addr.sin6_port = htons(23);
+  source_addr.sin6_addr = in6addr_loopback;
+  auto ep = to_ip_endpoint(reinterpret_cast<sockaddr_storage&>(source_addr));
+  auto dest_addr = to_sockaddr(ep);
+  CAF_CHECK_EQUAL(memcmp(&source_addr, &dest_addr, sizeof(sockaddr_in6)), 0);
```
Neverlord: Is bitwise equality really guaranteed without calling `memset` on the structs beforehand?

jakobod: I thought that initialising a struct like this: `sockaddr_in6 source_addr = {};` would initialize all members to their default values. With that in mind bitwise equality should be guaranteed without calling `memset`. Is that wrong?

Neverlord: Seems like [you're right](https://en.cppreference.com/w/c/language/struct_initialization):

> All members that are not initialized explicitly are initialized implicitly the same way as objects that have static storage duration.

TIL. 👍

PR #22, opened 2019-09-04
Add function to resolve local addresses and interface names

josephnoir:
The lookup of the any addresses is currently hard coded. We could redirect to resolve for these provided something [like this](https://github.com/actor-framework/incubator/commit/aa2bd7b9efcde2e143f2f8a5ef8ed3cc85212bd3) is acceptable.

### DIFF
--- a/libcaf_net/caf/net/ip.hpp
+++ b/libcaf_net/caf/net/ip.hpp
@@ -30,6 +30,10 @@ namespace ip {
 /// Returns all IP addresses of to `host` (if any).
 std::vector<ip_address> resolve(string_view host);
 
+/// Returns the IP addresses for a local endpoint, which is either an address,
+/// an interface, or localhost.
+std::vector<ip_address> local_addresses(string_view host);
+
 /// Returns the hostname of this device.
 std::string hostname();
 
--- a/libcaf_net/caf/net/ip.hpp
+++ b/libcaf_net/caf/net/ip.hpp
@@ -30,9 +30,15 @@ namespace ip {
 /// Returns all IP addresses of to `host` (if any).
 std::vector<ip_address> resolve(string_view host);
 
+/// Returns all IP addresses of to `host` (if any).
+std::vector<ip_address> resolve(ip_address host);
+
 /// Returns the IP addresses for a local endpoint, which is either an address,
-/// an interface, or localhost.
+/// an interface name, or the string "localhost".
 std::vector<ip_address> local_addresses(string_view host);
+
+/// Returns the IP addresses for a local endpoint address.
+std::vector<ip_address> local_addresses(ip_address host);
 
 /// Returns the hostname of this device.
 std::string hostname();
--- a/libcaf_net/caf/net/ip.hpp
+++ b/libcaf_net/caf/net/ip.hpp
@@ -27,10 +27,10 @@ namespace caf {
 namespace net {
 namespace ip {
 
-/// Returns all IP addresses of to `host` (if any).
+/// Returns all IP addresses of `host` (if any).
 std::vector<ip_address> resolve(string_view host);
 
-/// Returns all IP addresses of to `host` (if any).
+/// Returns all IP addresses of `host` (if any).
 std::vector<ip_address> resolve(ip_address host);
 
 /// Returns the IP addresses for a local endpoint, which is either an address,
--- a/libcaf_net/src/ip.cpp
+++ b/libcaf_net/src/ip.cpp
@@ -58,10 +58,10 @@ namespace ip {
 namespace {
 
 // Dummy port to resolve empty string with getaddrinfo.
-constexpr auto dummy_port = "42";
-constexpr auto v4_any_addr = "0.0.0.0";
-constexpr auto v6_any_addr = "::";
-constexpr auto localhost = "localhost";
+constexpr string_view dummy_port = "42";
+constexpr string_view v4_any_addr = "0.0.0.0";
+constexpr string_view v6_any_addr = "::";
+constexpr string_view localhost = "localhost";
 
 void* fetch_in_addr(int family, sockaddr* addr) {
   if (family == AF_INET)
@@ -115,7 +115,7 @@ std::vector<ip_address> resolve(string_view host) {
   addrinfo* tmp = nullptr;
   std::string host_str{host.begin(), host.end()};
   if (getaddrinfo(host.empty() ? nullptr : host_str.c_str(),
-                  host.empty() ? dummy_port : nullptr, &hint, &tmp)
+                  host.empty() ? dummy_port.data() : nullptr, &hint, &tmp)
       != 0)
     return {};
   std::unique_ptr<addrinfo, decltype(freeaddrinfo)*> addrs{tmp, freeaddrinfo};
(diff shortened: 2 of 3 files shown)


libcaf_net/src/ip.cpp
```diff
@@ -113,6 +146,114 @@ std::string hostname() {
...
+
+std::vector<ip_address> local_addresses(string_view host) {
+  using adapters_ptr = std::unique_ptr<IP_ADAPTER_ADDRESSES, void (*)(void*)>;
+  using wbuf_ptr = std::unique_ptr<IP_ADAPTER_ADDRESSES, void (*)(void*)>;
+  if (!host.empty() && host.compare(v4_any_addr) == 0)
+    return {ip_address{make_ipv4_address(0, 0, 0, 0)}};
+  if (!host.empty() && host.compare(v6_any_addr) == 0)
+    return {ip_address{}};
+  ULONG len = 0;
+  auto err = GetAdaptersAddresses(AF_UNSPEC, GAA_FLAG_INCLUDE_PREFIX, nullptr,
```
Neverlord: The variable is used only once.

```suggestion
  if (GetAdaptersAddresses(AF_UNSPEC, GAA_FLAG_INCLUDE_PREFIX, nullptr, nullptr, &len) != ERROR_BUFFER_OVERFLOW)
```

libcaf_net/src/ip.cpp
```diff
@@ -113,6 +146,114 @@ std::string hostname() {
...
+  }
+  auto adapters = adapters_ptr{reinterpret_cast<IP_ADAPTER_ADDRESSES*>(
+                                 ::new uint8_t[len]),
+                               free};
+  if (!adapters) {
+    CAF_LOG_ERROR("malloc failed");
+    return {};
+  }
+  // TODO: The API example propopses to try three times.
+  err = GetAdaptersAddresses(AF_UNSPEC, GAA_FLAG_INCLUDE_PREFIX, nullptr,
```
Neverlord: Re-using variables makes the code hard to follow. Also, we can instead simply wrap the call into an `if` without the variable.

libcaf_net/src/ip.cpp
```diff
@@ -113,6 +146,114 @@ std::string hostname() {
...
+                             adapters.get(), &len);
+  if (err != ERROR_SUCCESS) {
+    CAF_LOG_ERROR("failed to get adapter addresses");
+    return {};
+  }
+  char ip_buf[INET6_ADDRSTRLEN];
+  char name_buf[HOST_NAME_MAX];
+  std::vector<std::pair<std::string, ip_address>> interfaces;
```
Neverlord: Why are we filling up this vector instead of doing the work right there in the for loop?

I think if we move the loop into a template function, then we can safe a lot of unnecessary work and also make the code more modular:

```
template <class F>
void for_each_adapter(IP_ADAPTER_ADDRESSES* first, F f) {
  for (; first != nullptr; first = first->Next) {
    // ...
  }
}

// Call side.
if (host.empty()) {
  auto f = [&](string_view, ip_address addr) {
    results.push_back(addr);
  };
  for_each_adapter(adapters.get(), f);
} else if ...
```

libcaf_net/src/ip.cpp
```diff
@@ -113,6 +146,114 @@ std::string hostname() {
...
+  else if (auto err = parse(host, host_addr))
+    find_by_name(interfaces, host, results);
+  else
+    find_by_addr(interfaces, host_addr, results);
+  return results;
+}
+
+#else // CAF_WINDOWS
+
+std::vector<ip_address> local_addresses(string_view host) {
```
Neverlord: Same comments as in the Windows implementation.

libcaf_net/test/ip.cpp
```diff
@@ -30,16 +30,48 @@
...
+
+CAF_TEST_FIXTURE_SCOPE(ip_tests, fixture)
 
 CAF_TEST(resolve) {
-  ip_address v4_local{make_ipv4_address(127, 0, 0, 1)};
-  ip_address v6_local{{0}, {0x1}};
-  auto addrs = ip::resolve("localhost");
+  addrs = ip::resolve("localhost");
+  CAF_CHECK(!addrs.empty());
+  CAF_CHECK(contains(v4_local) || contains(v6_local));
```
Neverlord: Doesn't this fail on systems without IPv6 enabled?

libcaf_net/test/ip.cpp
```diff
@@ -30,16 +30,48 @@
...
+}
+
+CAF_TEST(local addresses) {
+  CAF_MESSAGE("check: v4");
```
Neverlord: Why not separate each of these checks into an own TEST instead? That would make it more readable IMO.

libcaf_net/test/ip.cpp
josephnoir: I hope the `||` will prevent that. It will fail if there is neither v4 nor v6.

libcaf_net/test/ip.cpp
josephnoir: Hm, I think the problem here is actually that the `local_address` call is wrong in returning an address when `::` is passed and the system doesn't have IPv6. Might be better to delegate calls with "localhost" or one of the "any" addresses to `resolve` which could handle this with #23.

libcaf_net/caf/net/ip.hpp
```diff
@@ -30,6 +30,10 @@ namespace ip {
 /// Returns all IP addresses of to `host` (if any).
 std::vector<ip_address> resolve(string_view host);
 
+/// Returns the IP addresses for a local endpoint, which is either an address,
+/// an interface, or localhost.
```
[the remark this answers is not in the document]
josephnoir: The function is meant to be used when creating local endpoints while resolve should be used for remote endpoints. That is why I would prefer an interface that mirrors the one of resolve.

Now we should answer the question if it would make sense for both functions to have  overloads for IP addresses and hosts.

* For `resolve` this would mean converting the ip address to a string because that is what `getaddrinfo` expects. 
* For `local_addresses` we could check first if the string is actually an IP address and dispatch that to the other overload. The result would be that both function call slightly different filters.

Having the overloads could save a back and forth conversion between IP address and string. Sounds like plan.

libcaf_net/caf/net/ip.hpp
```diff
@@ -30,6 +30,16 @@ namespace ip {
 /// Returns all IP addresses of to `host` (if any).
 std::vector<ip_address> resolve(string_view host);
 
+/// Returns all IP addresses of to `host` (if any).
```
Neverlord:
```suggestion
/// Returns all IP addresses of `host` (if any).
```

libcaf_net/caf/net/ip.hpp
```diff
@@ -30,6 +30,16 @@ namespace ip {
 /// Returns all IP addresses of to `host` (if any).
```
Neverlord:
```suggestion
/// Returns all IP addresses of `host` (if any).
```

libcaf_net/src/ip.cpp
```diff
@@ -106,6 +193,56 @@ std::vector<ip_address> resolve(string_view host) {
...
+  } else {
+    return local_addresses(host_ip);
+  }
+  return results;
+}
+
+std::vector<ip_address> local_addresses(ip_address host) {
+  auto v6_any = ip_address{{0}, {0}};
+  auto v4_any = ip_address{make_ipv4_address(0, 0, 0, 0)};
+  // TODO: If is any addr, call resolve with PR #23.
```
Neverlord: Are you planning to address this in this PR or separately? I'm fine with doing it later, but then please open an issue for the remaining work.

josephnoir: Tried to addresses that but it seems I screwed up somewhere. I'll take a look at it tomorrow.

josephnoir: Didn't think the last commit would fix the problem, but apparently it did.

PR #24, opened 2019-09-18
Add scaffold for new BASP implementation

Neverlord:
Note: BASP is *not* feature complete yet. However, the diff has already grown quite large and also includes refactoring of transport, etc.

I think it's best to merge this first chunk of BASP and then add the missing pieces in smaller PRs later. @jakobod, @josephnoir I've added you both for review, because I think it's important that we're on the same page at least for the changes to non-BASP classes. Since @josephnoir is familiar with the previous BASP implementation, I think it's ok if @jakobod only skims the implementation details of the `basp::application`. 🙂


libcaf_net/src/ec.cpp
```diff
@@ -0,0 +1,57 @@
...
+const char* ec_names[] = {
+  "none",
+  "invalid_magic_number",
+  "unexpected_number_of_bytes",
+  "unexpected_payload",
+  "missing_payload",
+  "illegal_state",
+  "invalid_handshake",
+  "missing_handshake",
+  "version_mismatch",
```
josephnoir: I think `"unexpected handshake"` is missing in the list.

libcaf_net/src/message_type.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "caf/string_view.hpp"
+
+namespace caf {
+namespace net {
+namespace basp {
+
+namespace {
+
+string_view message_type_names[] = {
```
josephnoir: Similar enums above (`ec`, `connection_state`) use a `const char*` for this list instead of a `string_view`.

libcaf_net/caf/net/basp/header.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+  header& operator=(const header&) noexcept = default;
+
+  int compare(header other) const noexcept;
+
+  /// @pre `bytes.size() == header_size`
+  static header from_bytes(span<const byte> bytes);
+};
+
+/// Size of a BASP header in serialized form
```
josephnoir:
```suggestion
/// Size of a BASP header in serialized form.
```

libcaf_net/src/application.cpp
```diff
@@ -0,0 +1,287 @@
...
+  serializer_impl<buffer_type> sink{system(), buf_};
+  if (auto err = sink(path)) {
+    CAF_LOG_ERROR("unable to serialize path");
+    return;
+  }
+  auto req_id = next_request_id_++;
+  auto hdr = to_bytes(header{message_type::resolve_request,
+                             static_cast<uint32_t>(buf_.size()), req_id});
+  if (auto err = write_packet(hdr, buf_)) {
+    CAF_LOG_ERROR("unable to serialize path");
```
josephnoir:
```suggestion
    CAF_LOG_ERROR("unable to write resolve request header");
```

libcaf_net/src/application.cpp
```diff
@@ -0,0 +1,287 @@
...
+    return err;
+  if (aid == 0) {
+    i->second.deliver(strong_actor_ptr{nullptr}, std::move(ifs));
+    return none;
+  }
+  i->second.deliver(proxies_->get_or_put(peer_id_, aid), std::move(ifs));
+  return none;
+}
+
+error application::generate_handshake() {
```
josephnoir: Not sure if we ever want to retransmit handshakes. If so, it might be a good idea to clear `buf_` before filling it.

libcaf_net/caf/net/basp/message_type.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+/// @addtogroup BASP
+
+/// Describes the first header field of a BASP message and determines the
+/// interpretation of the other header fields.
+enum class message_type : uint8_t {
+  /// Sends supported BASP version and node information to the server.
+  ///
+  /// ![](client_handshake.png)
+  handshake = 0x00,
```
jakobod: Is there a specific reason you are using hexadecimal here instead of decimal like you do in the other enum definitions?

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -61,18 +66,35 @@ class stream_transport {
...
+    return *this;
+  }
+
+  endpoint_manager& manager() {
+    return *manager_;
+  }
+
   // -- member functions -------------------------------------------------------
 
   template <class Parent>
```
jakobod: Is the `Parent` template and parameter still needed even though you the getters for `endpoint_managers` and so on above?

jakobod: Especially if you are using the specific type `endpoint_manager*` on line 237. This template seems especially odd..

libcaf_net/caf/net/endpoint_manager_impl.hpp
```diff
@@ -48,10 +52,18 @@ class endpoint_manager_impl : public endpoint_manager {
...
 
+  application_type& application() {
+    return transport_.application();
+  }
+
   transport_type& transport() {
     return transport_;
   }
 
+  endpoint_manager_impl& manager() {
```
jakobod: What is this getter for?

libcaf_net/caf/net/transport_worker.hpp
```diff
@@ -65,20 +81,16 @@ class transport_worker {
     application_.write_message(decorator, std::move(msg));
   }
 
-  template <class Parent>
-  void write_packet(Parent& parent, span<const byte> header,
```
jakobod: Why did you delete this function? Does the application need to get the `id_type` and pass it on to the transport itself?

libcaf_net/caf/net/basp/application.hpp
```diff
@@ -0,0 +1,217 @@
...
+  template <class Parent>
+  error write_message(Parent& parent,
+                      std::unique_ptr<endpoint_manager::message> ptr) {
+    // TODO: avoid extra copy of the payload
+    buf_.clear();
+    serializer_impl<buffer_type> sink{system(), buf_};
+    const auto& src = ptr->msg->sender;
+    const auto& dst = ptr->receiver;
+    if (dst == nullptr) {
+      // TODO: valid?
```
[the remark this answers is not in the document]
Neverlord: If we have a need for system messages then this could be one way of sending them.

libcaf_net/caf/net/basp/connection_state.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace net {
+namespace basp {
+
+/// @addtogroup BASP
+
+/// Stores the state of a connection in a `basp::application`.
+enum class connection_state {
+  /// Indicates that we successfully checked the magic number and now wait for
+  /// the handshake header.
+  await_handshake_header,
```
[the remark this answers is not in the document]
Neverlord: Sorry for the confusion. There is no magic number. I've started implementing BASP with a magic number but it only makes things complicated.

libcaf_net/src/application.cpp
```diff
@@ -0,0 +1,287 @@
...
+    case message_type::resolve_response:
+      return handle_resolve_response(write_packet, hdr, payload);
+    case message_type::heartbeat:
+      return none;
+    default:
+      return ec::unimplemented;
+  }
+}
+
+error application::handle_handshake(write_packet_callback&, header hdr,
```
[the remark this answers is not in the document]
Neverlord: Good catch!

libcaf_net/src/message_type.cpp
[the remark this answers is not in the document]
Neverlord: I think we should actually get rid of all the `const char*` stuff and make consistent use of `string_view` instead.

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -82,8 +104,7 @@ class stream_transport {
       CAF_LOG_DEBUG(CAF_ARG(len) << CAF_ARG(handle_.id) << CAF_ARG(*num_bytes));
       collected_ += *num_bytes;
       if (collected_ >= read_threshold_) {
-        auto decorator = make_write_packet_decorator(*this, parent);
-        worker_.handle_data(decorator, read_buf_);
+        worker_.handle_data(*this, read_buf_);
```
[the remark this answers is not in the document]
Neverlord: The decorator always calls `set_timeout` directly on the parent. Do you have a use case where we need to delegate `set_timeout` via decorator?

libcaf_net/src/application.cpp
```diff
@@ -0,0 +1,287 @@
...
+                                    byte_span payload) {
```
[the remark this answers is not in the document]
Neverlord: > I don't think this is used.

Is is now. 🙂

libcaf_net/caf/net/basp/message_type.hpp
Neverlord: Not really. Switched to decimal.

libcaf_net/caf/net/stream_transport.hpp
Neverlord: Yes, because `Parent` is the actual type, i.e., `endpoint_manager_impl<...>` and we might want to wrap the transport one day.

libcaf_net/caf/net/endpoint_manager_impl.hpp
Neverlord: We need to have a reference/pointer to the manager for creating proxies.

libcaf_net/caf/net/transport_worker.hpp
Neverlord: The only job of the decorator is to add this ID. That's why I changed `write_packet` in the decorator to avoid having all worker classes implement virtually identical `write_packet` functions. Here's the new code in the decorator:

```c++
  template <class... Ts>
  void write_packet(span<const byte> header, span<const byte> payload,
                    Ts&&... xs) {
    parent_.write_packet(header, payload, std::forward<Ts>(xs)...,
                         object_.id());
  }
```

libcaf_net/caf/net/stream_transport.hpp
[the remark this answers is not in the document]
josephnoir: I'm good with that. Just had it in the back of my head as a reason why we originally did this ...  that might have been a few iterations ago. ^^

libcaf_net/src/message_type.cpp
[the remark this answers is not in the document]
josephnoir: Let's do that.

libcaf_net/src/ec.cpp
```diff
@@ -0,0 +1,60 @@
...
+  "invalid_magic_number",
+  "unexpected_number_of_bytes",
+  "unexpected_payload",
+  "missing_payload",
+  "illegal_state",
+  "invalid_handshake",
+  "missing_handshake",
+  "unexpected_handshake",
+  "version_mismatch",
+  "unimplemented = 10",
```
josephnoir:
```suggestion
  "unimplemented",
```

Neverlord: I sneaked that in there to test your vigilance. 😏

libcaf_net/caf/net/stream_transport.hpp
jakobod: Oh yes, I see that! 👍

libcaf_net/caf/net/transport_worker.hpp
jakobod: This does indeed seem cleaner. 👍

josephnoir: Constant vigilance! 🧙‍♂️

josephnoir:
Ok, one last thing. How does the separate case or connection state `await_handshake_payload` help us? From the code it seems that we could do without it and move to `await_payload` directly. `handle` (in line 145) could then dispatch to `handle_handshake`. The difference might be that `connection_state::await_handshake_payload` does not check the payload size of the header.

Neverlord:
> Ok, one last thing. How does the separate case or connection state await_handshake_payload help us?

This state allows us to dispatch to `application::handle_handshake` only *once* and raise an error for repeated handshakes otherwise. Without that state, we can only dispatch on `hdr.type` and maybe do some guess work (check whether `peer_id_` is default-constructed or something along those lines). However, having this additional state gives us a very clean FSM that runs `handle_handshake` exactly once at the beginning and prohibits handshakes after the connection is established.

josephnoir:
Ah, because the case for `connection_state::await_header` simply processes the header of a repeated handshake, but handling the payload would then fail.

PR #25, opened 2019-09-19
Simplify socket design

Neverlord:
When I designed the sockets initially, I though there might be diamonds or other cross-relations in the socket hierarchy. However, this turned out not to be the case, so we can use a simpler design that doesn't require writing conversion operators.


libcaf_net/caf/net/socket.hpp
```diff
@@ -24,20 +24,55 @@
...
 };
 
+/// @relates socket
+template <class Inspector>
+typename Inspector::result_type inspect(Inspector& f, socket& x) {
+  return f(x.id);
+}
+
+/// @relates socket
+constexpr bool operator==(socket x, socket y) {
```
jakobod: Why not use `const socket&` for these and the following parameters?

libcaf_net/caf/net/socket.hpp
```diff
@@ -24,20 +24,55 @@
...
-  using super::super;
+  constexpr socket() : id(invalid_socket_id) {
+    // nop
+  }
+
+  constexpr explicit socket(socket_id id) : id(id) {
+    // nop
+  }
+
+  constexpr socket(const socket& other) : id(other.id) {
```
jakobod: Wouldn't `= default;` be enough for this trivial copy-constructor?

Neverlord: Absolutely. 🙂

libcaf_net/caf/net/socket.hpp
Neverlord: There is no reason to pass small objects by reference. Passing by value is both less noisy and can be more efficient. However, defining the comparison operators by hand is nonsense, we have `detail::comparable` for that.

jakobod: Even better. Thanks for the explanation! 👍

PR #27, opened 2019-10-04
Fix udp datagram socket

jakobod:
This PR fixes the `udp_datagram_socket` test. It also fixes a bug in `convert_ip_endpoint` and changes lookup of the local address to us the new `local_addresses()` function.

### DIFF
--- a/libcaf_net/src/udp_datagram_socket.cpp
+++ b/libcaf_net/src/udp_datagram_socket.cpp
@@ -100,8 +100,10 @@ variant<std::pair<size_t, ip_endpoint>, sec> read(udp_datagram_socket x,
                          << CAF_ARG(buf.size()) << " of " << CAF_ARG(num_bytes)
                          << " bytes");
     ip_endpoint ep;
-    if (auto err = detail::convert(addr, ep))
+    if (auto err = detail::convert(addr, ep)) {
+      CAF_ASSERT(err.category() == atom("system"));
       return static_cast<sec>(err.code());
+    }
     return std::pair<size_t, ip_endpoint>(*num_bytes, ep);
   } else {
     return get<sec>(ret);
--- a/libcaf_net/src/udp_datagram_socket.cpp
+++ b/libcaf_net/src/udp_datagram_socket.cpp
@@ -100,9 +100,8 @@ variant<std::pair<size_t, ip_endpoint>, sec> read(udp_datagram_socket x,
                          << CAF_ARG(buf.size()) << " of " << CAF_ARG(num_bytes)
                          << " bytes");
     ip_endpoint ep;
-    // TODO: how to properly pass error further?
     if (auto err = detail::convert(addr, ep))
-      return sec::runtime_error;
+      return static_cast<sec>(err.code());
     return std::pair<size_t, ip_endpoint>(*num_bytes, ep);
   } else {
     return get<sec>(ret);
--- a/libcaf_net/test/udp_datagram_socket.cpp
+++ b/libcaf_net/test/udp_datagram_socket.cpp
@@ -77,13 +77,17 @@ struct fixture : host_fixture {
     else
       CAF_FAIL("write returned an error: " << sys.render(get<sec>(write_ret)));
 
-    auto read_ret = read(receive_socket, make_span(buf));
-    if (auto read_res = get_if<std::pair<size_t, ip_endpoint>>(&read_ret)) {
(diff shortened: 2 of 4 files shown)


libcaf_net/src/udp_datagram_socket.cpp
```diff
@@ -100,9 +100,8 @@ variant<std::pair<size_t, ip_endpoint>, sec> read(udp_datagram_socket x,
                          << CAF_ARG(buf.size()) << " of " << CAF_ARG(num_bytes)
                          << " bytes");
     ip_endpoint ep;
-    // TODO: how to properly pass error further?
     if (auto err = detail::convert(addr, ep))
-      return sec::runtime_error;
```
Neverlord: I think we should add `CAF_ASSERT(err.category == atom("system"))` here to make sure the `static_cast` below is safe.

libcaf_net/test/udp_datagram_socket.cpp
```diff
@@ -30,85 +30,68 @@
...
-  if (contains(v4_local)) {
-    test_send_receive(v4_local);
-  } else if (contains(v6_local)) {
-    test_send_receive(v6_local);
-  } else {
-    CAF_FAIL("could not resolve 'localhost'");
-  }
+  std::vector<byte> buf(1024);
+  if (auto err = nonblocking(socket_cast<net::socket>(receive_socket), true))
+    CAF_FAIL("nonblocking returned an error" << err);
```
Neverlord:
```suggestion
    CAF_FAIL("setting socket to nonblocking failed: " << err);
```

libcaf_net/test/udp_datagram_socket.cpp
```diff
@@ -30,85 +30,68 @@
...
+  variant<std::pair<size_t, ip_endpoint>, sec> read_ret;
+  do {
+    read_ret = read(receive_socket, make_span(buf));
+    if (auto read_res = get_if<std::pair<size_t, ip_endpoint>>(&read_ret)) {
+      CAF_CHECK_EQUAL(read_res->first, hello_test.size());
+      buf.resize(read_res->first);
+    } else if (get<sec>(read_ret) != sec::unavailable_or_would_block) {
+      CAF_FAIL("read returned an error: " << sys.render(get<sec>(read_ret)));
+    }
+    CAF_CHECK_LESS(++rounds, 100);
```
Neverlord:
```suggestion
    if (++rounds == 100)
      CAF_FAIL("unable to receive data from the UDP socket");
```

libcaf_net/test/udp_datagram_socket.cpp
```diff
@@ -30,85 +30,68 @@
...
+  } while (holds_alternative<sec>(read_ret)
```
Neverlord: We call `CAF_FAIL` for any other error reason. What we actually want here is to wait until we have a valid result, I think this version is easier to comprehend:

```suggestion
  } while (read_ret.index() != 0);
```

libcaf_net/test/udp_datagram_socket.cpp
```diff
@@ -30,85 +30,68 @@
...
-  }
+  std::vector<byte> buf(1024);
+  if (auto err = nonblocking(socket_cast<net::socket>(receive_socket), true))
+    CAF_FAIL("nonblocking returned an error" << err);
+  CAF_CHECK_EQUAL(read(receive_socket, make_span(buf)),
+                  sec::unavailable_or_would_block);
+  CAF_MESSAGE("sending data to " << to_string(ep));
+  CAF_CHECK_EQUAL(write(send_socket, as_bytes(make_span(hello_test)), ep),
+                  hello_test.size());
+  int rounds = 0;
```
Neverlord: It's unclear from the context what a "round" is supposed to be here IMO. I think renaming the variable makes the intent clearer:

```suggestion
  int receive_attempts = 0;
```

libcaf_net/test/udp_datagram_socket.cpp
jakobod: What exactly does `index()` do?

Neverlord: It asks the `variant` [which type it contains](https://en.cppreference.com/w/cpp/utility/variant/index). Here, 0 is the index of the pair. Using `holds_alternative` is also OK, but only if we give the pair some typedef to reduce verbosity. :)

PR #29, opened 2019-10-08
No copy refactoring

jakobod:
This PR refactors the current implementation of the network stack to use actual `std::vector<byte>` instead of `span<byte>`. This allows moving data around instead of copying it from layer to layer.


jakobod:
I'm not sure about a couple of things. Could you take a look at them @Neverlord ? I will leave comments at the corresponding lines.

libcaf_net/caf/net/basp/application.hpp
```diff
@@ -73,7 +74,8 @@ class application {
       return err;
     auto hdr = to_bytes(header{message_type::handshake,
                                static_cast<uint32_t>(buf_.size()), version});
-    parent.write_packet(hdr, buf_);
+    std::vector<byte> header(hdr.begin(), hdr.end());
```
jakobod: Since `to_bytes` returns a `std::array<byte, header_size>` I had to copy the content.
Is there some deeper thought in using std::array other than the fixed size of `header` or can I refactor that function too?

libcaf_net/src/stream_socket.cpp
```diff
@@ -222,6 +222,10 @@ variant<size_t, sec> write(stream_socket x,
 
 variant<size_t, sec>
 check_stream_socket_io_res(std::make_signed<size_t>::type res) {
```
jakobod: `send` and `receive` return values mean different things. I think it would be a good idea to add a flag `rw` to this check function to distinguish between a `send` or a `receive` result and handle them separately.

Neverlord:
I think we covered all points in the f2f discussion today. After meging https://github.com/actor-framework/incubator/pull/28, I think you should rebase this branch with current `master`.

PR #32, opened 2019-10-11
No copy refactoring

jakobod:
This PR adds recycling of buffers to transports and refactors the stack so that no unnecessary copies are made while writing packets.

### DIFF
--- a/libcaf_net/caf/net/defaults.hpp
+++ b/libcaf_net/caf/net/defaults.hpp
@@ -1,0 +1,36 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright 2011-2018 Dominik Charousset                                     *
+ *                                                                            *
+ * Distributed under the terms and conditions of the BSD 3-Clause License or  *
+ * (at your option) under the terms and conditions of the Boost Software      *
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#pragma once
+
+#include <cstddef>
+
+// -- hard-coded default values for various CAF options ------------------------
+
+namespace caf {
+namespace defaults {
+
+namespace middleman {
+
+extern const size_t max_output_buffers;
+extern const size_t max_header_buffers;
+
+} // namespace middleman
+
+} // namespace defaults
+} // namespace caf
--- a/libcaf_net/caf/net/defaults.hpp
+++ b/libcaf_net/caf/net/defaults.hpp
@@ -27,7 +27,9 @@ namespace defaults {
 
 namespace middleman {
 
+/// Maximum number of payload buffers to keep in transport
 extern const size_t max_output_buffers;
+/// Maximum number of header buffers to keep in transport
 extern const size_t max_header_buffers;
 
 } // namespace middleman
--- a/libcaf_net/caf/net/defaults.hpp
+++ b/libcaf_net/caf/net/defaults.hpp
@@ -27,9 +27,10 @@ namespace defaults {
 
 namespace middleman {
 
-/// Maximum number of payload buffers to keep in transport
-extern const size_t max_output_buffers;
-/// Maximum number of header buffers to keep in transport
+/// Maximum number of cached buffers for sending payloads.
+extern const size_t max_payload_buffers;
+
+/// Maximum number of cached buffers for sending headers.
 extern const size_t max_header_buffers;
 
 } // namespace middleman
--- a/libcaf_net/caf/net/packet_writer.hpp
+++ b/libcaf_net/caf/net/packet_writer.hpp
@@ -32,18 +32,18 @@ class packet_writer {
 public:
   using buffer_type = std::vector<byte>;
 
-  virtual ~packet_writer() = default;
+  virtual ~packet_writer();
 
-  /// Returns the next header_buffer from transport
+  /// Returns a buffer for writing header information.
   virtual buffer_type next_header_buffer() = 0;
 
-  /// Returns the next payload_buffer from transport
-  virtual buffer_type next_buffer() = 0;
+  /// Returns a buffer for writing payload content.
+  virtual buffer_type next_payload_buffer() = 0;
 
   /// Convenience function to write a packet consisting of multiple buffers.
-  /// @param buffers the buffers that make up the packet. first buffer should
-  /// contain the header.
-  /// @warning takes ownership of `buffers`.
+  /// @param buffers all buffers for the packet. The first buffer is a header
+  ///                buffer, the other buffers are payload buffer.
+  /// @warning this function takes ownership of `buffers`.
   template <class... Ts>
   void write_packet(Ts&... buffers) {
     buffer_type* bufs[] = {&buffers...};
@@ -51,8 +51,8 @@ public:
   }
 
 protected:
-  /// Actual write_packet implementation.
-  /// @param buffers `span` containing all buffers of a packet.
+  /// Implementing function for `write_packet`.
+  /// @param buffers a `span` containing all buffers of a packet.
   virtual void write_impl(span<buffer_type*> buffers) = 0;
 };
 
--- a/libcaf_net/caf/net/packet_writer.hpp
+++ b/libcaf_net/caf/net/packet_writer.hpp
@@ -42,7 +42,7 @@ public:
   // Following buffers are one or more payload buffers for this packet.
   /// @warning takes ownership of `buffers`.
   template <class... Ts>
-  void write_packet(Ts... buffers) {
+  void write_packet(Ts&... buffers) {
     buffer_type* bufs[] = {&buffers...};
     write_impl(make_span(bufs, sizeof...(Ts)));
   }
--- a/libcaf_net/caf/net/packet_writer.hpp
+++ b/libcaf_net/caf/net/packet_writer.hpp
@@ -1,0 +1,55 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright 2011-2019 Dominik Charousset                                     *
+ *                                                                            *
+ * Distributed under the terms and conditions of the BSD 3-Clause License or  *
+ * (at your option) under the terms and conditions of the Boost Software      *
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#pragma once
+
+#include <vector>
+
+#include "caf/byte.hpp"
+#include "caf/net/fwd.hpp"
+#include "caf/span.hpp"
+
+namespace caf {
+namespace net {
+
+class packet_writer {
+public:
+  using buffer_type = std::vector<byte>;
+
+  // virtual ~packet_writer() = default;
+
+  virtual buffer_type next_header_buffer() = 0;
+
+  virtual buffer_type next_buffer() = 0;
+
+  // TODO: some documentation
+  // First buffer is header.
+  // Following buffers are one or more payload buffers for this packet.
+  /// @warning takes ownership of `buffers`.
+  template <class... Ts>
+  void write_packet(Ts... buffers) {
+    buffer_type* bufs[] = {&buffers...};
+    write_impl(make_span(bufs, sizeof...(Ts)));
+  }
+
+protected:
+  virtual void write_impl(span<buffer_type*> buffers) = 0;
+};
+
+} // namespace net
+} // namespace caf
--- a/libcaf_net/caf/net/packet_writer.hpp
+++ b/libcaf_net/caf/net/packet_writer.hpp
@@ -27,19 +27,22 @@
 namespace caf {
 namespace net {
 
+/// Implements an interface for packet writing in application-layers.
 class packet_writer {
 public:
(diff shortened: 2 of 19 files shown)


libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
 private:
   // -- private member functions -----------------------------------------------
 
   bool write_some() {
-    if (write_buf_.empty())
+    auto begin = [&]() { return write_queue_.begin(); };
+    // helper to sort empty buffers back into the right queues
+    auto recycle = [&]() {
+      auto is_header = [](std::pair<bool, buffer_type>& p) { return p.first; };
+      begin()->second.clear();
```
Neverlord:
```suggestion
      auto& front = write_queue_.front();
      bool is_header = front.first;
      auto& buf = front.second;
      buf.clear();
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
+      if (is_header(*begin()))
```
Neverlord:
```suggestion
      if (is_header)
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
+        free_header_bufs_.emplace_back(std::move(begin()->second));
```
Neverlord:
```suggestion
        free_header_bufs_.emplace_back(std::move(buf));
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
+      else
+        free_bufs_.emplace_back(std::move(begin()->second));
```
Neverlord:
```suggestion
        free_bufs_.emplace_back(std::move(buf));
```

Also, this misses a check against some upper bound. Both `free_bufs_` and `free_header_bufs_` can grow indefinitely at the moment.

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
-    if (auto num_bytes = get_if<size_t>(&ret)) {
-      CAF_LOG_DEBUG(CAF_ARG(len) << CAF_ARG(handle_.id) << CAF_ARG(*num_bytes));
-      // Update state.
-      written_ += *num_bytes;
-      if (written_ >= write_buf_.size()) {
-        written_ = 0;
-        write_buf_.clear();
-        return false;
+    do {
+      if (begin()->second.empty()) {
```
Neverlord:
```suggestion
      auto& front = write_queue.front();
      auto& buf = front.second;
      if (front.second.empty()) {
```

On a side note: empty buffers seem like a precondition violation. Can't we use `CAF_ASSERT(!buf.empty())` here instead?

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
-        worker_.handle_error(err);
+      // get size of send buffer
+      auto ret = send_buffer_size(handle_);
+      if (!ret) {
+        CAF_LOG_ERROR("send_buffer_size returned an error" << CAF_ARG(ret));
         return false;
       }
-    }
-    return true;
+      // is send buffer of socket full?
```
Neverlord:
```suggestion
      // Stop if the send buffer of the socket is full.
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
+      if (begin()->second.size() > *ret)
```
Neverlord:
```suggestion
      if (begin()->second.size() > *ret)
```

I think this check is wrong for two reasons:
- `ret` is the result of `getsockopt` with `SO_SNDBUF`, which returns the *maximum* socket send buffer size and *not* how much space is currently available
- What happens here if a payload is bigger than the socket buffer? That would block forever, no?

I don't think we need to keep track of `send_buffer_size` at all, actually. Rather than checking whether the entire buffer fits into the socket buffer, we should just write and deal with partial writes by storing where to continue writing the front buffer on the next invocation.

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +201,97 @@ class stream_transport {
...
+        return true;
+      CAF_LOG_TRACE(CAF_ARG(handle_.id));
+      auto write_ret = write(handle_, make_span(begin()->second));
+      if (auto num_bytes = get_if<size_t>(&write_ret)) {
+        CAF_LOG_DEBUG(CAF_ARG(handle_.id) << CAF_ARG(*num_bytes));
+        if (*num_bytes >= begin()->second.size())
```
Neverlord: ... `else`? The current logic could end up writing a partial buffer first and then repeating bytes later.

libcaf_net/caf/net/packet_writer_impl.hpp
```diff
@@ -18,16 +18,18 @@
...
 
 namespace caf {
 namespace net {
 
 /// Implements the interface for transport and application policies and
-/// dispatches member functions either to `decorator` or `parent`.
+/// dispatches member functions either to `object` or `parent`.
 template <class Object, class Parent>
-class write_packet_decorator {
+class packet_writer_impl final : public packet_writer {
```
Neverlord:
```suggestion
class packet_writer_decorator final : public packet_writer {
```

I think the `_impl` suffix here is misleading. There exist multiple implementations of `packet_writer`. This one simply decorates another `packet_writer`, so I think the `_decorator` suffix gives a better intuition.

libcaf_net/src/application.cpp
```diff
@@ -47,6 +48,81 @@ application::application(proxy_registry& proxies) : proxies_(proxies) {
   // nop
 }
 
+error application::write_message(
+  packet_writer& writer, std::unique_ptr<endpoint_manager_queue::message> ptr) {
+  CAF_ASSERT(ptr != nullptr);
+  CAF_ASSERT(ptr->msg != nullptr);
+  CAF_LOG_TRACE(CAF_ARG2("content", ptr->msg->content()));
+  auto payload_elem_buf = writer.next_buffer();
```
Neverlord:
```suggestion
  auto payload_prefix = writer.next_buffer();
```

Probably a better name. I don't have any intuition what `elem_buf` should mean.

libcaf_net/src/application.cpp
```diff
@@ -47,6 +48,81 @@ application::application(proxy_registry& proxies) : proxies_(proxies) {
...
+  if (src != nullptr) {
+    auto src_id = src->id();
+    system().registry().put(src_id, src);
+    if (auto err = sink(src->node(), src_id, dst->id(), ptr->msg->stages))
+      return err;
+  } else {
+    if (auto err = sink(node_id{}, actor_id{0}, dst->id(), ptr->msg->stages))
+      return err;
+  }
+  // TODO: Is this size correct?
```
Neverlord: Yes.

libcaf_net/src/application.cpp
```diff
@@ -47,6 +48,81 @@ application::application(proxy_registry& proxies) : proxies_(proxies) {
...
+  auto header_buf = writer.next_header_buffer();
+  to_bytes(header{message_type::actor_message,
+                  static_cast<uint32_t>(payload_elem_buf.size()
+                                        + ptr->payload.size()),
+                  ptr->msg->mid.integer_value()},
+           header_buf);
+  // TODO: OK to move payload out of message?
```
Neverlord: Yes. We destroy `ptr` at scope exit (and `ptr->payload` with it).

libcaf_net/src/header.cpp
```diff
@@ -46,17 +46,25 @@ header header::from_bytes(span<const byte> bytes) {
   return result;
 }
 
-std::array<byte, header_size> to_bytes(header x) {
-  std::array<byte, header_size> result;
-  auto ptr = result.data();
+void to_bytes_impl(const header& x, byte* ptr) {
```
Neverlord: Please put this into an anonymous namespace.

libcaf_net/caf/net/defaults.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace middleman {
+
+extern const size_t max_output_buffers;
+extern const size_t max_header_buffers;
+
+} // namespace middleman
+
+} // namespace defaults
+} // namespace caf
```
Neverlord: Missing newline at end of file.

libcaf_net/caf/net/defaults.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <cstddef>
+
+// -- hard-coded default values for various CAF options ------------------------
+
+namespace caf {
+namespace defaults {
+
+namespace middleman {
+
+extern const size_t max_output_buffers;
```
Neverlord: Please add documentation to the constants to explain what they actually configure.

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +207,101 @@ class stream_transport {
...
+    // helper to sort empty buffers back into the right queues
+    auto recycle = [&]() {
+      auto& front = write_queue_.front();
+      auto& is_header = front.first;
+      auto& buf = front.second;
+      written_ = 0;
+      buf.clear();
+      if (is_header && free_header_bufs_.size() < max_header_bufs_)
+        free_header_bufs_.emplace_back(std::move(buf));
+      else if (free_bufs_.size() < max_output_bufs_)
```
Neverlord: This is a bug. Now, we end up adding header buffers to `free_bufs_` if there's not enough room in `free_header_bufs_`:

```c++
      if (is_header) {
        if (free_header_bufs_.size() < max_header_bufs_)
          free_header_bufs_.emplace_back(std::move(buf));
      } else if (free_bufs_.size() < max_output_bufs_)
        free_bufs_.emplace_back(std::move(buf));
      }
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +207,101 @@ class stream_transport {
...
-      CAF_LOG_DEBUG(CAF_ARG(len) << CAF_ARG(handle_.id) << CAF_ARG(*num_bytes));
-      // Update state.
-      written_ += *num_bytes;
-      if (written_ >= write_buf_.size()) {
-        written_ = 0;
-        write_buf_.clear();
-        return false;
+    do {
+      auto& buf = write_queue_.front().second;
+      if (buf.empty()) {
```
Neverlord: Can we drop this branch and instead use `CAF_ASSERT(!buf.empty())`?

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +207,101 @@ class stream_transport {
...
+      auto data = buf.data() + written_;
+      auto len = buf.size() - written_;
+      auto write_ret = write(handle_, make_span(data, len));
+      if (auto num_bytes = get_if<size_t>(&write_ret)) {
+        CAF_LOG_DEBUG(CAF_ARG(handle_.id) << CAF_ARG(*num_bytes));
+        if (*num_bytes + written_ >= buf.size()) {
+          recycle();
+          written_ = 0;
+        } else {
+          written_ = *num_bytes;
```
Neverlord:
```suggestion
          written_ += *num_bytes;
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -252,7 +313,7 @@ class stream_transport {
   size_t written_;
 
   endpoint_manager* manager_;
-};
+}; // namespace net
```
Neverlord: How did that comment slip in here?

```suggestion
};
```

libcaf_net/src/application.cpp
```diff
@@ -257,23 +279,22 @@ error application::handle_actor_message(write_packet_callback&, header hdr,
...
+                                          byte_span received) {
+  CAF_LOG_TRACE(CAF_ARG(rec_hdr) << CAF_ARG2("received.size", received.size()));
+  CAF_ASSERT(rec_hdr.type == message_type::resolve_request);
   size_t path_size = 0;
-  binary_deserializer source{&executor_, payload};
+  binary_deserializer source{&executor_, received};
   if (auto err = source.begin_sequence(path_size))
     return err;
-  // We expect the payload to consist only of the path.
+  // We expect the received to consist only of the path.
```
Neverlord:
```suggestion
  // We expect the received buffer to contain the path only.
```

libcaf_net/caf/net/stream_transport.hpp
jakobod: Oh my.. :eyes:

libcaf_net/caf/net/stream_transport.hpp
jakobod: I don't think so. At least not with the current test-implementations. Some tests still send empty buffers I think.

jakobod: Should I add that change to this PR or can I move that to another one?

jakobod: Done.

libcaf_net/caf/net/packet_writer.hpp
```diff
@@ -0,0 +1,60 @@
...
+public:
+  using buffer_type = std::vector<byte>;
+
+  virtual ~packet_writer() = default;
+
+  /// Returns the next header_buffer from transport
+  virtual buffer_type next_header_buffer() = 0;
+
+  /// Returns the next payload_buffer from transport
+  virtual buffer_type next_buffer() = 0;
```
Neverlord: I think we should make the API a bit clearer:

```suggestion
  virtual buffer_type next_payload_buffer() = 0;
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -4,7 +4,7 @@
  *                     | |     / _ \ | |_       Actor                         *
  *                     | |___ / ___ \|  _|      Framework                     *
  *                      \____/_/   \_|_|                                      *
- *                                                                            *
+ *                                                                            *#
```
Neverlord:
```suggestion
 *                                                                            *
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -198,49 +207,100 @@ class stream_transport {
...
+    } while (!write_queue_.empty());
+    return false;
   }
 
   worker_type worker_;
   stream_socket handle_;
 
-  std::vector<byte> read_buf_;
-  std::vector<byte> write_buf_;
+  std::deque<buffer_type> free_header_bufs_;
```
Neverlord: I didn't catch that earlier, but using a `std::deque` here is overkill. We pay for a complex data structure without actually needing it. Also, a FIFO policy on buffer means that we are always touching cold memory. A LIFO policy is more efficient and also can use a more efficient data structure (good ol' `vector`). As an added bonus, we can get rid of the `max_...` members and simply ask the vector for its capacity.

PR #34, opened 2019-10-24
Add `datagram-transport`

jakobod:
This PR adds a `udp_datagram_transport` to the network stack.
It also adds some minor changes to the `udp_datagram_socket` and some more tests.


libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -0,0 +1,327 @@
...
+ ******************************************************************************/
+
+#pragma once
+
+#include <deque>
+#include <unordered_map>
+
+#include "caf/byte.hpp"
+#include "caf/detail/socket_sys_aliases.hpp"
+#include "caf/detail/socket_sys_includes.hpp"
```
Neverlord: These header should never get pulled into a CAF header. Especially `socket_sys_includes`, which pulls in Windows headers that can mess up downstream code. Why did you need them here? If something is missing from the interface for `udp_datagram_socket`?

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -0,0 +1,327 @@
...
+    return handle_;
+  }
+
+  actor_system& system() {
+    return manager().system();
+  }
+
+  application_type& application() {
+    // TODO: This wont work. We need information on which application is wanted
+    return application_type{};
```
Neverlord: Not only isn't this working, this returns a reference to a temporary and causes UB.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -33,25 +34,27 @@ namespace caf {
 namespace net {
 
 /// implements a dispatcher that dispatches between transport and workers.
```
Neverlord:
```suggestion
/// Implements a dispatcher that dispatches between transport and workers.
```

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -65,48 +68,48 @@ class transport_worker_dispatcher {
 
   template <class Parent>
   error handle_data(Parent& parent, span<byte> data, id_type id) {
-    auto it = workers_by_id_.find(id);
-    if (it == workers_by_id_.end()) {
-      // TODO: where to get node_id from here?
+    auto worker = find_by_id(id);
+    if (!worker) {
+      // TODO: where to get id_type from here?
```
Neverlord: What is the TODO about? We do have an ID inside this function.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -65,48 +68,48 @@ class transport_worker_dispatcher {
...
       add_new_worker(parent, node_id{}, id);
-      it = workers_by_id_.find(id);
+      worker = find_by_id(id);
```
Neverlord: We perform a lookup right after inserting the worker. `add_new_worker` should simply return the new worker.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -65,48 +68,48 @@ class transport_worker_dispatcher {
...
-    auto it = workers_by_node_.find(nid);
-    if (it == workers_by_node_.end()) {
+    auto nid = receiver->node();
+    auto worker = find_by_node(nid);
+    if (!worker) {
       // TODO: where to get id_type from here?
       add_new_worker(parent, nid, id_type{});
-      it = workers_by_node_.find(nid);
+      worker = find_by_node(nid);
     }
```
Neverlord: Stay DRY: this duplicates code from `handle_data` and should thus go into a function.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -65,48 +68,48 @@ class transport_worker_dispatcher {
...
     worker->write_message(parent, std::move(msg));
   }
 
   template <class Parent>
-  void write_packet(Parent& parent, span<const byte> header,
-                    span<const byte> payload, id_type id) {
-    parent.write_packet(header, payload, id);
+  void resolve(Parent& parent, const uri& locator, const actor& listener) {
+    if (auto worker = find_by_node(make_node_id(locator)))
+      worker->resolve(parent, locator.path(), listener);
```
Neverlord: This lacks an else tree. `listener` will wait forever on a message on error.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -140,13 +143,36 @@ class transport_worker_dispatcher {
   }
 
 private:
+  worker_ptr find_by_node(const node_id& nid) {
+    if (workers_by_node_.empty())
+      return nullptr;
+    auto it = workers_by_node_.find(nid);
+    if (it == workers_by_node_.end()) {
+      CAF_LOG_ERROR("could not find worker by node: " << CAF_ARG(nid));
```
Neverlord:
```suggestion
      CAF_LOG_DEBUG("could not find worker by node: " << CAF_ARG(nid));
```

I wouldn't log an error here. This can simply happen after losing a connection and does not impede program correctness.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -140,13 +143,36 @@ class transport_worker_dispatcher {
...
+      return nullptr;
+    }
+    return it->second;
+  }
+
+  worker_ptr find_by_id(const IdType& id) {
```
Neverlord:
```suggestion
  worker_ptr find_worker(const id_type& id) {
```

Same reasoning as above.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
```diff
@@ -140,13 +143,36 @@ class transport_worker_dispatcher {
   }
 
 private:
+  worker_ptr find_by_node(const node_id& nid) {
```
Neverlord:
```suggestion
  worker_ptr find_worker(const node_id& nid) {
```

Find by node... what? The signature makes it clear what our key is, so the function name should reflect what the function is actually doing.

libcaf_net/caf/net/udp_datagram_socket.hpp
```diff
@@ -56,9 +56,23 @@ error allow_connreset(udp_datagram_socket x, bool new_value);
 variant<std::pair<size_t, ip_endpoint>, sec> read(udp_datagram_socket x,
                                                   span<byte> buf);
 
+/// Sends the content of `bufs` as a datagram to the endpoint `ep` on socket
+/// `x`.
+/// @param x x The UDP socket for sending datagrams.
```
Neverlord:
```suggestion
/// @param x The UDP socket for sending datagrams.
```

libcaf_net/caf/net/udp_datagram_socket.hpp
```diff
@@ -56,9 +56,23 @@ error allow_connreset(udp_datagram_socket x, bool new_value);
...
+/// @param bufs Points to the datagram to send, scattered across up to 10
+///             buffers.
+/// @param ep The enpoint to send the datagram to.
+/// @returns The number of written bytes on success, otherwise an error code.
+/// @relates udp_datagram_socket
+/// @post either the result is a `sec` or a positive (non-zero) integer
```
Neverlord:
```suggestion
```

This adds nothing. The type `variant<size_t, sec>` already says that and `@returns` already documents what the integer is there for.

libcaf_net/test/datagram_transport.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  using buffer_ptr = std::shared_ptr<buffer_type>;
+
+  fixture() : shared_buf(std::make_shared<buffer_type>(1024)) {
+    mpx = std::make_shared<multiplexer>();
+    if (auto err = mpx->init())
+      CAF_FAIL("mpx->init failed: " << sys.render(err));
+    mpx->set_thread_id();
+    CAF_CHECK_EQUAL(mpx->num_socket_managers(), 1u);
+    if (auto err = parse("127.0.0.1:0", ep))
```
Neverlord: Hardcoding an IPv4 address won't work on systems such as OpenBSD. We also had trouble on other systems already with this since it's configuration-dependent whether this works if the UDP socket prefers IPv6. Can we ask the UDP socket for its local address instead?

libcaf_net/test/datagram_transport.cpp
```diff
@@ -0,0 +1,261 @@
...
+    recv_socket = receive_pair.first;
+    ep.port(htons(receive_pair.second));
+    CAF_MESSAGE("sending message to " << CAF_ARG(ep));
+    if (auto err = nonblocking(recv_socket, true))
+      CAF_FAIL("nonblocking() returned an error: " << err);
+  }
+
+  ~fixture() {
+    close(send_socket);
+    close(recv_socket);
```
Neverlord: These two sockets should go into a socket guard instead. Can be a followup, since we just extended the socket guard API in #35 to allow this.

libcaf_net/test/datagram_transport.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  using buffer_ptr = std::shared_ptr<buffer_type>;
+
+public:
+  dummy_application(buffer_ptr rec_buf)
+    : rec_buf_(std::move(rec_buf)){
+      // nop
+    };
+
+  ~dummy_application() = default;
```
Neverlord:
```suggestion
```

There's no need to spell that out.

libcaf_net/test/datagram_transport.cpp
```diff
@@ -0,0 +1,261 @@
...
+  udp_datagram_socket recv_socket;
+};
+
+class dummy_application {
+  using buffer_type = std::vector<byte>;
+
+  using buffer_ptr = std::shared_ptr<buffer_type>;
+
+public:
+  dummy_application(buffer_ptr rec_buf)
```
Neverlord:
```suggestion
  explicit dummy_application(buffer_ptr rec_buf)
```

Even though it's "just" a unit test: please don't allow the compiler to turn any buffer it sees into an application. 😉

libcaf_net/test/transport_worker_dispatcher.cpp
```diff
@@ -197,13 +199,14 @@ struct fixture : host_fixture {
     auto p = make_actor<dummy_actor, strong_actor_ptr>(aid, nid, &sys, cfg);
     auto test_span = as_bytes(make_span(hello_test));
     buffer_type payload(test_span.begin(), test_span.end());
-    auto strong_actor = actor_cast<strong_actor_ptr>(p);
+    auto receiver = actor_cast<strong_actor_ptr>(p);
+    if (!receiver)
+      CAF_FAIL("receiver cast failed");
```
Neverlord:
```suggestion
      CAF_FAIL("failed to cast receiver to a strong_actor_ptr");
```

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -0,0 +1,327 @@
...
+  using application_type = typename Factory::application_type;
+
+  // -- constructors, destructors, and assignment operators --------------------
+
+  datagram_transport(udp_datagram_socket handle, factory_type factory)
+    : dispatcher_(*this, std::move(factory)),
+      handle_(handle),
+      max_consecutive_reads_(0),
+      read_threshold_(1024),
+      max_(1024),
```
Neverlord: Is this a hard maximum? In any case, I think the magic numbers should at least go into some header with a proper name.

libcaf_net/caf/net/datagram_transport.hpp
jakobod: The forward declaration of `udp_datagram_socket` was missing in `net::fwd.hpp`. I fixed that issue already in the quic-branch.

jakobod: These must have slipped into the file somehow.. I don't need them.

libcaf_net/caf/net/datagram_transport.hpp
jakobod: I see that, but I needed a way to make the compiler happy. 
Problem is, that we will need some identifying component in this call. The dispatcher must find the proper worker before it can return the corresponding application. Do you have an idea how to fix that?

libcaf_net/caf/net/transport_worker_dispatcher.hpp
jakobod: This was a typo. I actually meant `node_id`.
When there is no worker for incoming traffic, the dispatcher instantiates a new one. For that it needs the `node_id` and the `id_type` of the incoming traffic.

libcaf_net/caf/net/transport_worker_dispatcher.hpp
jakobod: There is no way to return an error from `resolve` - But I can add that.

Neverlord: I don't mean *returning* an error. I mean sending an error to the listener. :)

libcaf_net/caf/net/transport_worker_dispatcher.hpp
jakobod: The problem is, that I need to distinguish between `node_id` and `id_type`. I refactored the code now, but I don't see how not to duplicate this code. Could you have a look at the changed parts?

libcaf_net/caf/net/transport_worker_dispatcher.hpp
jakobod: How exactly though? I would need to add return type `error` to the whole chain. But what could I do about the `endpoint_manager_impl`?

jakobod: You answered faster than I was typing.. ignore that last comment!

libcaf_net/caf/net/datagram_transport.hpp
jakobod: I have removed this call now.

PR #35, opened 2019-10-25
Streamline stream_transport test

jakobod:
This PR streamlines the `stream_transport` test-suite.

### DIFF
--- a/libcaf_net/test/stream_transport.cpp
+++ b/libcaf_net/test/stream_transport.cpp
@@ -55,9 +55,9 @@ struct fixture : test_coordinator_fixture<>, host_fixture {
     mpx->set_thread_id();
     CAF_CHECK_EQUAL(mpx->num_socket_managers(), 1u);
     auto sockets = unbox(make_stream_socket_pair());
-    send_socket.reset(sockets.first);
-    recv_socket.reset(sockets.second);
-    if (auto err = nonblocking(*recv_socket, true))
+    send_socket_guard.reset(sockets.first);
+    recv_socket_guard.reset(sockets.second);
+    if (auto err = nonblocking(recv_socket_guard.socket(), true))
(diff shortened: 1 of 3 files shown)


libcaf_net/test/stream_transport.cpp
```diff
@@ -39,27 +40,48 @@ using namespace caf;
...
+
+  using buffer_ptr = std::shared_ptr<buffer_type>;
+
+  fixture() : recv_buf(1024), shared_buf{std::make_shared<buffer_type>()} {
     mpx = std::make_shared<multiplexer>();
     if (auto err = mpx->init())
       CAF_FAIL("mpx->init failed: " << sys.render(err));
     mpx->set_thread_id();
+    CAF_CHECK_EQUAL(mpx->num_socket_managers(), 1u);
+    auto sockets = unbox(make_stream_socket_pair());
```
Neverlord: I like putting the buffers into the fixture. However, putting the sockets into the fixture makes ownership a bit fuzzy. The fixture takes ownership of one socket, but it just assumes that an endpoint manager is going to take ownership for the second one.

I'd either move this code to a function (and have the fixture not participating in socket lifetimes) or put both sockets into a socket guard. Then, the endpoint manager can explicitly take ownership of it via `release`. If you go with the socket guard route: it currently has no accessor for the socket, but I see no reason not to add `operator*`, `operator->` and `get()` to `socket_guard` in order to give it some smart pointer like functionality.

jakobod: Sounds like a good idea. :)

Neverlord: Looking at the code, the smart-pointer-ish interface is more confusing then helpful. I've refactored it again to have an interface that makes more sense for a guard type. :)

PR #38, opened 2019-11-12
Fix basp::application test

jakobod:
This PR fixes the `basp::application` test which was broken by a recent change in `response_promise`.


Neverlord:
Thanks for fixing! 👍

PR #39, opened 2019-11-12
Move to C++17

jakobod:
This PR updates the incubator build harness to use C++17. It also updates the Jenkinsfile to run the same stages as the actor-framework repository does.


Neverlord:
Thanks for taking care of this! 🙏🏻

PR #41, opened 2019-11-13
Add transport_base class

### LINKED ISSUE #36: Code duplication in different transport implementations
While going through the different transport implementations I noticed that we currently duplicate a lot of code.
I think we can clean up those implementations by supplying something like a `transport_base`, which encapsulates the shared state and functions. The different transport-implementations would only need to implement their individual `init()`, `handle_read_event()` and `handle_write_event()` functions + special state/functions.

Would that be a good idea? Or is there something that speaks against such a change @Neverlord ?

jakobod:
This PR adds a base class for transport implementations.

Closes #36

### DIFF
--- a/libcaf_net/caf/net/datagram_transport.hpp
+++ b/libcaf_net/caf/net/datagram_transport.hpp
@@ -19,22 +19,17 @@
 #pragma once
 
 #include <deque>
-#include <unordered_map>
+#include <vector>
 
-#include "caf/byte.hpp"
-#include "caf/error.hpp"
 #include "caf/fwd.hpp"
 #include "caf/logger.hpp"
-#include "caf/net/defaults.hpp"
 #include "caf/net/endpoint_manager.hpp"
 #include "caf/net/fwd.hpp"
-#include "caf/net/receive_policy.hpp"
 #include "caf/net/transport_base.hpp"
 #include "caf/net/transport_worker_dispatcher.hpp"
 #include "caf/net/udp_datagram_socket.hpp"
 #include "caf/sec.hpp"
 #include "caf/span.hpp"
-#include "caf/variant.hpp"
 
 namespace caf::net {
 
@@ -76,6 +71,7 @@ public:
   // -- public member functions ------------------------------------------------
 
   error init(endpoint_manager& manager) override {
+    CAF_LOG_TRACE("");
     if (auto err = super::init(manager))
       return err;
     prepare_next_read();
@@ -87,6 +83,7 @@ public:
     auto ret = read(this->handle_, make_span(this->read_buf_));
     if (auto res = get_if<std::pair<size_t, ip_endpoint>>(&ret)) {
       auto num_bytes = res->first;
+      CAF_LOG_DEBUG("received " << num_bytes << " bytes");
       auto ep = res->second;
       this->read_buf_.resize(num_bytes);
       this->next_layer_.handle_data(*this, make_span(this->read_buf_),
@@ -115,6 +112,7 @@ public:
     return write_some();
   }
 
+  // TODO: remove this function. `resolve` should add workers when needed.
   error add_new_worker(node_id node, id_type id) {
     auto worker = this->next_layer_.add_new_worker(*this, node, id);
     if (!worker)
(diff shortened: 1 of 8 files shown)


libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -19,193 +19,118 @@
...
-
-  void set_timeout(uint64_t timeout_id, id_type id) {
-    dispatcher_.set_timeout(timeout_id, id);
-  }
-
-  void handle_error(sec code) {
-    dispatcher_.handle_error(code);
-  }
-
+  // TODO: remove this function. `resolve` should add workers when needed.
```
josephnoir: Hm?

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -219,75 +144,66 @@ class datagram_transport {
...
       } else {
         auto err = get<sec>(write_ret);
         if (err != sec::unavailable_or_would_block) {
-          CAF_LOG_DEBUG("send failed" << CAF_ARG(err));
-          dispatcher_.handle_error(err);
+          CAF_LOG_ERROR("write failed" << CAF_ARG(err));
+          this->next_layer_.handle_error(err);
           return false;
         }
+        CAF_LOG_DEBUG("write returned `unavailable_or_would_block`");
```
josephnoir: Does this fit our usually style for return values?

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -18,248 +18,163 @@
...
   // -- member types -----------------------------------------------------------
 
   using application_type = Application;
 
-  using transport_type = stream_transport;
-
   using worker_type = transport_worker<application_type>;
 
-  using buffer_type = std::vector<byte>;
+  using id_type = unit_t;
```
josephnoir: Could be `super::id_type`.

libcaf_net/caf/net/transport_base.hpp
```diff
@@ -0,0 +1,182 @@
...
+    next_layer_.set_timeout(timeout_id, id);
+  }
+
+  void handle_error(sec code) {
+    next_layer_.handle_error(code);
+  }
+
+  // -- (pure) virtual functions -----------------------------------------------
+
+  virtual void configure_read(receive_policy::config){
```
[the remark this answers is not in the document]
jakobod: I tried. clang-format won't leave that space there. Probably an issue with the `.clang-format` file?

libcaf_net/caf/net/transport_base.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+  // -- buffer management ------------------------------------------------------
+
+  buffer_type next_header_buffer() {
+    return next_buffer_impl(header_bufs_);
+  }
+
+  buffer_type next_payload_buffer() {
+    return next_buffer_impl(payload_bufs_);
+  }
```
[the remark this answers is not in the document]
jakobod: Because we're trying to limit vector reallocations.`header_buffers` are always the same (small) size. If one of those `header_buffers` is passed as a `payload_buffer` it would have to be reallocated to fit the bigger payload size. Hence we're managing both buffer-types differently.

libcaf_net/caf/net/datagram_transport.hpp
jakobod: Adding new workers for new connections shouldn't happen manually. If a resolve_request succeeds a new `transport_worker` should be added automatically. This is a function needed for the current test implementation because `resolve` doesn't actually create workers automatically by now.. I will implement that and remove this function later on.

libcaf_net/caf/net/datagram_transport.hpp
jakobod: Not sure what you mean?

libcaf_net/caf/net/stream_transport.hpp
jakobod: :+1:

libcaf_net/caf/net/datagram_transport.hpp
josephnoir: I was talking about the backticks around the error. If we write the complete message ourselves we could just say what is happening instead of printing the name of the variable we use internally for this error code.

libcaf_net/caf/net/transport_base.hpp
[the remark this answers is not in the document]
josephnoir: Ah. I missed that you pass different arguments. Makes sense!

PR #44, opened 2019-11-19
Add all.hpp header

jakobod:
This PR adds the `all.hpp` header so a `FindCAF_NET.cmake` script can find the built library.

### DIFF
--- a/libcaf_net/caf/net/all.hpp
+++ b/libcaf_net/caf/net/all.hpp
@@ -51,17 +51,3 @@
 #include "caf/net/transport_worker.hpp"
 #include "caf/net/transport_worker_dispatcher.hpp"
 #include "caf/net/udp_datagram_socket.hpp"
-
-#include "caf/net/backend/test.hpp"
-
-#include "caf/net/basp/application.hpp"
-#include "caf/net/basp/connection_state.hpp"
-#include "caf/net/basp/constants.hpp"
-#include "caf/net/basp/ec.hpp"
-#include "caf/net/basp/header.hpp"
-#include "caf/net/basp/message_queue.hpp"
-#include "caf/net/basp/message_type.hpp"
-#include "caf/net/basp/remote_message_handler.hpp"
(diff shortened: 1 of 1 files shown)


libcaf_net/caf/net/all.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#pragma once
+
+#include "caf/net/actor_proxy_impl.hpp"
```
josephnoir: Sounds like it should be private.

libcaf_net/caf/net/all.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "caf/net/datagram_socket.hpp"
+#include "caf/net/datagram_transport.hpp"
+#include "caf/net/defaults.hpp"
+#include "caf/net/endpoint_manager.hpp"
+#include "caf/net/endpoint_manager_impl.hpp"
```
josephnoir: This one as well, maybe?

Neverlord:
> @Neverlord, what do you think?

I think the all header should contain everything in `caf::net` but nothing from nested namespaces such as `net::backend`, `net::basp`, etc. We should probably start auto-generating the `all.hpp` headers.

If the `_impl` classes shouldn't be part of the public API then the also shouldn't live in `caf::net` but in a nested namespace (or `caf::detail`).

Neverlord:
> `FindCAF_NET.cmake` script can find the built library.

One final thought on that: our findCAF script should pick up `caf_net` just fine, e.g., `find_package(CAF COMPONENTS core net)`

PR #45, opened 2019-11-19
Refactor nested to concatenated namespaces

jakobod:
This PR refactors all occurrences of nested to concatenated namespaces.


cmake/get_compiler_version.cpp
```diff
@@ -2,16 +2,12 @@
 using namespace std;
 
 int main() {
-#   ifdef __clang__
```
Neverlord: This file is obsolete. Please remove.

jakobod: Done.

PR #46, opened 2019-11-22
Add Doorman

jakobod:
This PR adds a `doorman` which listens for and accepts incoming TCP connections.


libcaf_net/caf/net/doorman.hpp
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  template <class Parent>
+  bool handle_read_event(Parent& parent) {
+    auto x = net::accept(acceptor_);
+    if (!x) {
+      CAF_LOG_ERROR("accept failed:" << parent.system().render(x.error()));
+      return false;
+    }
+    // TODO this won't work anymore
```
Neverlord: What is this comment all about?

libcaf_net/caf/net/socket_manager.hpp
```diff
@@ -49,6 +49,12 @@ class socket_manager : public ref_counted {
     return handle_;
   }
 
+  /// Returns a pointer to the multiplexer running the socket manager or
+  /// `nullptr` if the manager is no assigned to one.
```
Neverlord: That sentence feels a bit unwieldy. Isn't it just trying to say something along the lines of `Returns the parent multiplexer`?

libcaf_net/src/tcp_accept_socket.cpp
```diff
@@ -118,7 +118,7 @@ make_tcp_accept_socket(const uri::authority_type& node, bool reuse_addr) {
   if (auto ip = get_if<ip_address>(&node.host))
     return make_tcp_accept_socket(ip_endpoint{*ip, node.port}, reuse_addr);
   auto host = get<std::string>(node.host);
-  auto addrs = ip::resolve(host);
+  auto addrs = ip::resolve(host); // TODO: switch to local_addresses()
```
Neverlord: Why add a TODO instead of doing the switch?

libcaf_net/test/doorman.cpp
```diff
@@ -0,0 +1,174 @@
...
+ * Distributed under the terms and conditions of the BSD 3-Clause License or  *
+ * (at your option) under the terms and conditions of the Boost Software      *
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#define CAF_SUITE doorman
```
Neverlord:
```suggestion
#define CAF_SUITE net.doorman
```

Also, the file should land in `libcaf_net/test/net/doorman.cpp`

libcaf_net/test/doorman.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+CAF_TEST(tcp connect) {
+  auto acceptor = unbox(make_tcp_accept_socket(auth, false));
+  auto port = unbox(local_port(socket_cast<network_socket>(acceptor)));
+  auto acceptor_guard = make_socket_guard(acceptor);
+  CAF_MESSAGE("opened acceptor on port " << port);
+  uri::authority_type dst;
+  dst.port = port;
+  dst.host = "localhost"s;
+  auto conn = make_socket_guard(unbox(make_connected_tcp_stream_socket(dst)));
```
Neverlord: Is there really no better way to test this than making localhost connections?

libcaf_net/caf/net/doorman.hpp
jakobod: I actually don't know anymore.. I mean it works  👀

libcaf_net/caf/net/socket_manager.hpp
jakobod: I think so too. Maybe @josephnoir knows more than we do?

libcaf_net/src/tcp_accept_socket.cpp
jakobod: I forgot about that.

libcaf_net/test/doorman.cpp
jakobod: I actually don't think there is. Do you think one could try to use unix sockets to test such a routine?

libcaf_net/caf/net/socket_manager.hpp
josephnoir: Nope, he doesn't.

libcaf_net/test/doorman.cpp
jakobod: I have split up the two test suites to cleanup the test scaffold. 
I also discussed the doorman test with @josephnoir and we came to the conclusion that:
1. splitting up the tests would be a good idea.
2. Testing tcp accept functionality without actual tcp would not be a good idea.
3. The doorman test could be realized without a localhost connection, but would be pretty hacky.. (correct me if I'm wrong Joseph.).

jakobod:
I have fixed the pointed out comment and cleaned the test suite a bit. Could you take another look @Neverlord ?

libcaf_net/test/doorman.cpp
```diff
@@ -0,0 +1,161 @@
...
+  CAF_CHECK_EQUAL(mgr->init(), none);
+  auto before = mpx->num_socket_managers();
+  CAF_CHECK_EQUAL(before, 2u);
+  uri::authority_type dst;
+  dst.port = port;
+  dst.host = "localhost"s;
+  CAF_MESSAGE("connecting to doorman on: " << dst);
+  auto conn = make_socket_guard(unbox(make_connected_tcp_stream_socket(dst)));
+  CAF_MESSAGE("waiting for connection");
+  while (mpx->num_socket_managers() != before + 1)
```
Neverlord: This needs an additional exit condition, because otherwise the loop can run forever. I go ahead and merge, but this needs a followup.

jakobod: Will provide that tomorrow! :+1:

PR #47, opened 2019-11-25
Refactor code to use new binary_serializer

jakobod:
This PR refactors the incubator code base to use the new `binary_serializer`.

### DIFF
--- a/libcaf_net/caf/net/stream_transport.hpp
+++ b/libcaf_net/caf/net/stream_transport.hpp
@@ -85,8 +85,8 @@ public:
       this->collected_ += *num_bytes;
       if (this->collected_ >= this->read_threshold_) {
         if (auto err = this->next_layer_.handle_data(*this,
-                                                     as_bytes(make_span(
-                                                       this->read_buf_)))) {
+                                                     make_span(
+                                                       this->read_buf_))) {
           CAF_LOG_ERROR("handle_data failed: " << CAF_ARG(err));
           return false;
         }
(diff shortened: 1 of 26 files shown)


jakobod:
Two things:
1. I had to include `message.hpp` and `make_message.hpp` in two places because `make_error` would not  compile without those headers. This seems like somewhat of a hacky solution.. 
2. The `caf::span` type is not implicitly casted from `span<T>` to `span<const T>` anymore. Hence I had to add lots of `as_bytes(make_span(x))` calls. This seems unwieldy..

Neverlord:
> This seems like somewhat of a hacky solution..

I think working with forward declarations whenever possible is generally a good thing. Sure, it adds some "inconvenience" to the call side by having to include a couple extra headers, but I think it's worth it.

> his seems unwieldy..

Agree. Please fix this upstream rather than working around it multiple times. 🙂

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -84,7 +84,9 @@ class stream_transport : public stream_transport_base<Application> {
                     << CAF_ARG(this->handle_.id) << CAF_ARG(*num_bytes));
       this->collected_ += *num_bytes;
       if (this->collected_ >= this->read_threshold_) {
-        if (auto err = this->next_layer_.handle_data(*this, this->read_buf_)) {
+        if (auto err = this->next_layer_.handle_data(*this,
+                                                     make_span(
```
Neverlord: You've removed the `as_bytes`, but this should also compile without `make_span`, does it not?

jakobod: It should compile without `make_span` but isn't it clearer like this?

Neverlord: I don't think so. 😉

The function takes a span, which is just a view into *some* sort of buffer. Just like we wouldn't manually convert a string to a `string_view`, this only adds extra noise and verbosity.

jakobod: Alright, I get that. I will remove those 👍

jakobod: Done.

PR #51, opened 2020-01-08
Implement proper shutdown of multiplexer

jakobod:
This PR adds a proper implementation for shutting down the `net` network stack.

The PR already contains #50 . So that should be merged first.


libcaf_net/src/multiplexer.cpp
```diff
@@ -242,8 +242,10 @@ void multiplexer::shutdown() {
       for (size_t i = 1; i < managers_.size(); ++i)
         write_to_pipe(4, managers_[i]);
     }
+    return true;
   } else {
     write_to_pipe(5, nullptr);
+    return false;
```
Neverlord: Why not run `run()` here and return `void` instead?

jakobod: I thought calling `run()` from the `middleman`at the same place as `mpx_->join()` would be clearer code.

Neverlord: A shutdown function that returns a boolean that has nothing to do with the success of the operation is not particularly clear to me TBH. 😉

jakobod: I see that and have changed it. :+1: Just out of curiosity: Would this be passable solution if there was type-information in form of an enum or something instead of the bool?

jakobod: Okay wait. @josephnoir just had something to say about this. Your proposed solution would not work. We have to differentiate the two cases and call `run()` from the `middleman` in order to safely shutdown the `multiplexer. I fixed this  and have a commit on the way.

jakobod: Okay wait. @josephnoir just had something to say about this. Your proposed solution would not work. We have to differentiate the two cases and call `run()` from the `middleman` in order to safely shutdown the `multiplexer. I fixed this  and have a commit on the way.

libcaf_net/caf/net/multiplexer.hpp
```diff
@@ -119,6 +135,9 @@ class multiplexer : public std::enable_shared_from_this<multiplexer> {
 
   /// Guards `write_handle_`.
   std::mutex write_lock_;
+
+  /// Signals shutdown has been requested.
+  bool will_shutdown_;
```
Neverlord:
```suggestion
  bool shutting_down_ = false;
```

libcaf_net/caf/net/multiplexer.hpp
```diff
@@ -61,6 +61,8 @@ class multiplexer : public std::enable_shared_from_this<multiplexer> {
   /// Returns the index of `mgr` in the pollset or `-1`.
   ptrdiff_t index_of(const socket_manager_ptr& mgr);
 
+  bool is_same_thread();
```
Neverlord:
```suggestion

```

libcaf_net/src/multiplexer.cpp
```diff
@@ -222,6 +234,21 @@ void multiplexer::run() {
...
 
+void multiplexer::shutdown() {
+  if (std::this_thread::get_id() == tid_) {
+    will_shutdown_ = true;
+    if (managers_.size() == 1) {
+      close_pipe();
+    } else {
+      // First manager is the pollset_updater. Skip it and delete later.
+      for (size_t i = 1; i < managers_.size(); ++i)
+        write_to_pipe(4, managers_[i]);
```
Neverlord:
```c++
shutting_down_ = true;
if (managers_.size() > 1) {
  // iterate managers and stop reading (remove if not writing)
}
close_pipe();
```

libcaf_net/caf/net/multiplexer.hpp
```diff
@@ -71,6 +73,10 @@ class multiplexer : public std::enable_shared_from_this<multiplexer> {
   /// @thread-safe
   void register_writing(const socket_manager_ptr& mgr);
 
+  /// Unregisters `mgr` for read and write events.
+  /// @thread-safe
+  void unregister_manager(const socket_manager_ptr& mgr);
```
Neverlord:
```suggestion

```

libcaf_net/src/multiplexer.cpp
```diff
@@ -105,6 +104,10 @@ ptrdiff_t multiplexer::index_of(const socket_manager_ptr& mgr) {
   return i == last ? -1 : std::distance(first, i);
 }
 
+bool multiplexer::is_same_thread() {
+  return std::this_thread::get_id() == tid_;
+}
+
 void multiplexer::register_reading(const socket_manager_ptr& mgr) {
   if (std::this_thread::get_id() == tid_) {
     if (mgr->mask() != operation::none) {
```
Neverlord:
```c++
if (shutting_down_) {
  // discard
} else if (mgr->mask() != operation::none) {
```

Same in `register_writing`.

PR #52, opened 2020-01-08
Fix XCode CMake generator issue

jakobod:
This PR just pulls the fix for  the XCode CMake generator issue from master. 
See actor-framework/actor-framework#1000 for more information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,11 @@ include(CheckCSourceRuns)          # Check whether compiler produces binaries
 include(GNUInstallDirs)            # Sets default install paths
 include(GenerateExportHeader)      # Auto-generates dllexport macros
 
+
+# -- project-specific CMake settings -------------------------------------------
+
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
 # -- check whether we are running as CMake subdirectory ------------------------
 
 get_directory_property(_parent PARENT_DIRECTORY)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -206,20 +211,15 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL "${CMAKE_CURRENT_BINARY_DIR}")
...
-  set(LIBRARY_OUTPUT_PATH "${CMAKE_CURRENT_BINARY_DIR}/lib")
+if(NOT XCODE)
+  # set binary output path if not defined by user
+  if(NOT EXECUTABLE_OUTPUT_PATH)
+    set(EXECUTABLE_OUTPUT_PATH "${CMAKE_CURRENT_BINARY_DIR}/bin")
+  endif()
+  # set library output path if not defined by user
+  if(NOT LIBRARY_OUTPUT_PATH)
+    set(LIBRARY_OUTPUT_PATH "${CMAKE_CURRENT_BINARY_DIR}/lib")
+  endif()
```
Neverlord: Please remove these lines. This fails the coverage test now, because `caf-net-test` lands in `${CMAKE_CURRENT_BINARY_DIR}/bin`. This got copied over from the main repo, but here we didn't set `EXECUTABLE_OUTPUT_PATH` and `LIBRARY_OUTPUT_PATH` before. I think we shouldn't start doing it either. If anything, we should drop these lines from the main repo as well.

PR #57, opened 2020-01-16
Add nodelay opt

jakobod:
We forgot to set the `TCP_NODELAY` socket option and this PR fixes that.


Neverlord:
Does setting this option has any effect on performance?

jakobod:
Actually it does. The pingpong benchmark with `caf::net` performs equally / a bit better with this option than the benchmark with caf::io.
On the other hand the streaming benchmark is not affected at all.

PR #60, opened 2020-01-20
Refactor write events

jakobod:
This PR refactors the handling of write events in the `udp` and `tcp` transports.

### DIFF
--- a/libcaf_net/caf/net/datagram_transport.hpp
+++ b/libcaf_net/caf/net/datagram_transport.hpp
@@ -107,8 +107,10 @@ public:
       }
       return false;
     };
-    while (write_some() && fetch_next_message())
-      ; // nop
+    do {
+      if (!write_some())
+        return false;
+    } while (fetch_next_message());
     return !packet_queue_.empty();
   }
 
--- a/libcaf_net/caf/net/datagram_transport.hpp
+++ b/libcaf_net/caf/net/datagram_transport.hpp
@@ -98,17 +98,20 @@ public:
   }
 
   bool handle_write_event(endpoint_manager& manager) override {
-    CAF_LOG_TRACE(CAF_ARG(this->handle_.id)
-                  << CAF_ARG2("queue-size", packet_queue_.size()));
-    // Try to write leftover data.
-    write_some();
-    // Get new data from parent.
-    for (auto msg = manager.next_message(); msg != nullptr;
-         msg = manager.next_message()) {
-      this->next_layer_.write_message(*this, std::move(msg));
-    }
-    // Write prepared data.
-    return write_some();
+    CAF_LOG_TRACE(CAF_ARG2("handle", this->handle_.id)
+                  << CAF_ARG2("queue-size", this->packet_queue_.size()));
+    auto get_messages = [&] {
+      if (packet_queue_.empty()) {
+        for (auto msg = manager.next_message(); msg != nullptr;
+             msg = manager.next_message()) {
+          this->next_layer_.write_message(*this, std::move(msg));
+        }
+      }
+      return !packet_queue_.empty();
+    };
+    while (write_some() && get_messages())
+      ; // nop
+    return !packet_queue_.empty();
   }
 
   // TODO: remove this function. `resolve` should add workers when needed.
@@ -159,7 +162,6 @@ private:
   }
 
   bool write_some() {
-    CAF_LOG_TRACE(CAF_ARG2("handle", this->handle_.id));
     // Helper function to sort empty buffers back into the right caches.
     auto recycle = [&]() {
       auto& front = packet_queue_.front();
@@ -192,13 +194,11 @@ private:
         if (err != sec::unavailable_or_would_block) {
           CAF_LOG_ERROR("write failed" << CAF_ARG(err));
           this->next_layer_.handle_error(err);
-          return false;
         }
-        CAF_LOG_DEBUG("write returned `unavailable_or_would_block`");
-        return true;
+        return false;
       }
     }
-    return false;
+    return true;
   }
 
   std::deque<packet> packet_queue_;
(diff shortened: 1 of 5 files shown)


libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -104,18 +104,18 @@ class stream_transport : public stream_transport_base<Application> {
...
+                  << CAF_ARG2("queue-size", write_queue_.size()));
+    auto get_messages = [&] {
+      for (auto msg = manager.next_message(); msg != nullptr;
+           msg = manager.next_message())
+        this->next_layer_.write_message(*this, std::move(msg));
+      return !write_queue_.empty();
+    };
+    while (write_some() && get_messages())
+      ; // nop
+    return !write_queue_.empty();
```
Neverlord: This can fetch more messages then we can currently write. Wouldn't it make more sense to only get one message at a time and stop as soon as we can't write to the socket buffer any longer?

jakobod: That would be a good idea. I tried that aproach already, but currently we are not able to actually get to the point where we have more than one message to handle per event. So I thought we could stick with getting all messages.

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -98,17 +98,16 @@ class datagram_transport : public datagram_transport_base<Factory> {
...
-    for (auto msg = manager.next_message(); msg != nullptr;
-         msg = manager.next_message()) {
-      this->next_layer_.write_message(*this, std::move(msg));
-    }
-    // Write prepared data.
-    return write_some();
+    auto get_message = [&] {
+      if (auto msg = manager.next_message())
+        this->next_layer_.write_message(*this, std::move(msg));
+      return !packet_queue_.empty();
```
Neverlord: This returns `false` after dequeueing the last element.

```c++
      if (auto msg = manager.next_message()) {
        this->next_layer_.write_message(*this, std::move(msg));
        return true;
      }
      return false;
```

Also, I find the lambda name misleading. This doesn't just *get* a message. It fetches the next message and writes it. `fetch_next_message`?

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -98,17 +98,16 @@ class datagram_transport : public datagram_transport_base<Factory> {
...
+    };
+    while (write_some() && get_message())
```
Neverlord: Is `&&` really correct? Intuitively, I think we want to try getting the next message if `write_some` failed and stop once we couldn't make any progress, i.e., `write_some() || get_message()`.

libcaf_net/caf/net/datagram_transport.hpp
jakobod: > This returns false after dequeueing the last element.

Why though? Every message that is dequeued should eventually be moved into the packet_queue of the transport. Hence this will only return `false` if there is nothing left in the queue.
But I see that just returning true and false is more simple. :+1:

libcaf_net/caf/net/datagram_transport.hpp
jakobod: I think so. `write_some` returns `false` on error which means we don't want to get more data from our parent. If `write_some` returns `true` all queued messages have been sent hence we want to get more messages from our parent. With `||` this would not work that way. We would only stop getting more messages from our parent in the case that there are no more messages to get -- which is not what we want I guess.

libcaf_net/caf/net/datagram_transport.hpp
Neverlord: Ah, the `packet_queue_` is where `write_message` is supposed to put the outgoing messages? But you can still have the case that (for whatever reason) the next layer drops a message. In this case, we still want to return true to try again on the next cycle.

jakobod: > Ah, the packet_queue_ is where write_message is supposed to put the outgoing messages?

Exactly. 
I have changed it locally already. Your proposed change is definitely better suited.

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -98,17 +98,18 @@ class datagram_transport : public datagram_transport_base<Factory> {
...
+    auto fetch_next_message = [&] {
+      if (auto msg = manager.next_message()) {
+        this->next_layer_.write_message(*this, std::move(msg));
+        return true;
+      }
+      return false;
+    };
+    while (write_some() && fetch_next_message())
+      ; // nop
+    return !packet_queue_.empty();
```
Neverlord: Sorry, I think the logic on this is still off.

Previously, we returned the result of `write_some()`, which only returns `false` in case of an error. But now, this functions returns false if it has nothing more to write.

Doesn't the multiplexer remove this socket for write events when we return `false`? Looking at the code in `write_some()`, I get even more suspicious. The function writes as much as possible from `packet_queue_` *and returns `true`* if there is still data available to write but the socket just can't take any more data.

My understanding is that this loop now pushes even more data into the cache, while `write_some()` returns `true` all the while although the socket buffer is at capacity (paying one syscall per iteration). Once `manager.next_message()` returns `nullptr`, the loop stops. However, the functions now returns `false` even if there was no error. You tweaked the results or `write_some`, though. Isn't this silently discarding errors now?

My feeling is that the code wasn't really correct before (ignoring the result for the first `write_some`) but now became completely broken. Either my mental model of the new multiplexer is wrong or this code. 😕

jakobod: > Previously, we returned the result of write_some(), which only returns false in case of an error. But now, this functions returns false if it has nothing more to write.

This is not entirely true. `write_some` returned `false` in the case of an empty buffer too. 
I see the problem with the discarded errors though and have altered the code in my last commit. Now the transport should have a similar behavior, without discarding any errors.

PR #62, opened 2020-01-20
Add serializer worker

jakobod:
This PR adds workers for serializing outgoing messages.
With this change there are currently two similar types of workers. We should think about merging them and  removing the two similar implementations.

### DIFF
--- a/libcaf_net/CMakeLists.txt
+++ b/libcaf_net/CMakeLists.txt
@@ -44,6 +44,7 @@ set(CAF_NET_SOURCES
   src/defaults.cpp
   src/message_queue.cpp
   src/worker.cpp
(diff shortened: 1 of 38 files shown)


jakobod:
Since the `actor_proxy` is using cached `payload_buffers` now too, we ran into a race on the payload_buffer cache. I fixed it by synchronizing the access to it, but I am not sure if that is the best solution. What is more expensive synchronizing the access or allocating memory for the buffers? 
Would you mind taking a look into that @Neverlord ? I can run some benchmarks looking into this tomorrow.

Neverlord:
@jakobod I close this, since we've put the workers on hold for now.

PR #65, opened 2020-03-16
Port message-rework changes to incubator

jakobod:
This PR ports the incubator code to work with the upcoming changes of actor-framework/actor-framework#1057.
Since this change depends on those changes, it should be merged after that PR is approved.


jakobod:
Dang, I missed those! Will do right away :+1:

Neverlord:
Thanks! Obviously, I'll wait with the merge until https://github.com/actor-framework/actor-framework/pull/1057 landed in CAF. 🙂

jakobod:
Okay just for my sanity: I have fixed the build errors that were still present, but I'm a little bit confused about why this works. I have to have two `type_id` headers which have to be initialized using `caf::init_global_meta_objects<>()`, for one`incubator_test.hpp` and the other one is `bb_test_type_ids.hpp`. I initialize the incubator_test ones in the main I added, and the `bb_test` one in the constructor of the `bb_test_config`. Are the bb tests not run by the main?

cmake/incubator-test.cpp
```diff
@@ -1,11 +1,9 @@
...
 
-#include "caf/fwd.hpp"
-#include "caf/test/dsl.hpp"
-#include "caf/test/unit_test_impl.hpp"
-#include "caf/type_id.hpp"
+#include "incubator-test.hpp"
 
 int main(int argc, char** argv) {
-  caf::init_global_meta_objects<>();
+  caf::core::init_global_meta_objects();
```
Neverlord: For future reference: It's actually better to call this *after* the `init_global_meta_objects` below, because initializing the largest block first makes sure you allocate memory only once.

PR #71, opened 2020-06-03
Add tcp_backend

jakobod:
This PR adds a `tcp_backend` and `publish` and `remote_actor` functions to the `middleman`.
closes #58 .

PR #76, opened 2020-06-16
Add udp backend

### LINKED ISSUE #54: Implement Datagram layer for applications
Currently it is only possible to use stream oriented transport protocols to communicate between different nodes. If we want to be able to use datagram oriented protocols we will have to implement some kind of datagram layer, which extracts the data and passes it piecewise into the application.

jakobod:
This PR adds a simple `udp backend` and a `datagram_adaptor`, which splits received datagrams into header and payload which can then be passed separately to the following application.

Also included are some minor bugfixes in the `datagram_transport` implementation including the `transport_worker_dispatcher`.

This PR is based on #71 and should be merged/reviewed after that has been merged.
closes #54

### DIFF
--- a/libcaf_net/caf/net/datagram_adaptor.hpp
+++ b/libcaf_net/caf/net/datagram_adaptor.hpp
@@ -28,20 +28,55 @@ template <class Application>
 class datagram_adaptor {
 public:
   datagram_adaptor(Application application)
-    : application_(std::move(application)), missing_(0), passed_(0) {
+    : application_(std::move(application)) {
     // nop
   }
 
   template <class Parent>
-  error handle_data(Parent&, span<const byte> received) {
-    auto writer = make_packet_writer_decorator(*this, parent);
+  error init(Parent& parent) {
+    return application_.init(parent);
+  }
+
+  template <class Parent>
+  error write_message(Parent& parent,
+                      std::unique_ptr<endpoint_manager_queue::message> msg) {
+    return application_.write_message(parent, std::move(msg));
+  }
+
+  template <class Parent>
+  error handle_data(Parent& parent, span<const byte> received) {
     if (auto err = application_.handle_data(
-          writer_, make_span(received.data(), basp::header_size)))
+          parent, make_span(received.data(), basp::header_size)))
       return err;
-    if (auto err = application_.handle_data(
-          writer_, make_span(received.data() + basp::header_size,
-                             received.size() - basp::header_size))
-      return err;
+    auto data = received.data() + basp::header_size;
+    auto size = received.size() - basp::header_size;
+    if (size > 0)
+      return application_.handle_data(parent, make_span(data, size));
+    return none;
+  }
+
+  template <class Parent>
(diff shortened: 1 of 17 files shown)


libcaf_net/caf/net/defaults.hpp
```diff
@@ -36,4 +36,7 @@ CAF_NET_EXPORT extern const size_t max_header_buffers;
 /// Port to listen on for tcp.
 CAF_NET_EXPORT extern const uint16_t tcp_port;
 
+/// Port to listen on for udp.
+CAF_NET_EXPORT extern const uint16_t udp_port;
```
Neverlord, line 40: Do we need two constants? I assume this is for exposing BASP. Since we're going to use the locator (URI) as identifier for the CAF node, we must use either TCP or UDP. In other words, one default port would suffice.

libcaf_net/caf/net/datagram_adaptor.hpp
```diff
@@ -0,0 +1,86 @@
...
+  template <class Parent>
+  error write_message(Parent& parent,
+                      std::unique_ptr<endpoint_manager_queue::message> msg) {
+    return application_.write_message(parent, std::move(msg));
+  }
+
+  template <class Parent>
+  error handle_data(Parent& parent, span<const byte> received) {
+    if (auto err = application_.handle_data(
+          parent, make_span(received.data(), basp::header_size)))
```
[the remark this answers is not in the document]
jakobod: TBH I think with the proposal of #81 the `datagram_adaptor` would be obsolete. We should discuss that issue first before renaming or rethinking the `datagram_adaptor` for `BASP`.

libcaf_net/caf/net/defaults.hpp
jakobod: I thought that BASP could be used with `udp` or `tcp` at the same time, thus both transports require different ports to be accessible from the network.

Neverlord: In the new model, the node ID *is* the locator in order to get rid of the pseudo-DNS CAF is currently doing. This means there must be exactly one valid way to connect to a BASP instance.

jakobod: I forgot about the fixed node ID.. Still I am not sure if that is the best way to solve this problem. Is it really necessary to allow only one possible way to reach a node via `BASP`?

Neverlord: By definition, a node can only have a single ID. If we allow multiple IDs pointing to the same BASP, we have the exact same issues we wanted to get rid of with the redesign that @josephnoir initiated.

We could of course offer a "connect there and see who's responding" function, but it would convolute the design quite a bit. Because then, we can't use the ID any longer to identify connections. I honestly don't see an upside here that would be worth the troubles in the first place.

jakobod: I can understand that. I'll remove the duplicate port field when I'm fixing the other remark.

PR #85, opened 2020-07-27
Switch to more lightweight socket API wrapper

Neverlord:
The current API with `variant<size_t, sec>` has some nice semantic properties, but comes at a cost. This set of changes essentially reverts the socket API to a thin wrapper over the POSIX interface, but using `ptrdiff_t` instead of `ssize_t` (the latter is no C standard type).

### DIFF
--- a/libcaf_net/caf/net/stream_transport.hpp
+++ b/libcaf_net/caf/net/stream_transport.hpp
@@ -73,6 +73,11 @@ public:
 
   bool handle_read_event(endpoint_manager&) override {
     CAF_LOG_TRACE(CAF_ARG2("handle", this->handle().id));
+    auto fail = [this](sec err) {
+      CAF_LOG_DEBUG("read failed" << CAF_ARG(err));
+      this->next_layer_.handle_error(err);
+      return false;
+    };
     for (size_t reads = 0; reads < this->max_consecutive_reads_; ++reads) {
       auto buf = this->read_buf_.data() + collected_;
       size_t len = read_threshold_ - collected_;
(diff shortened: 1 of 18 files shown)


libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -74,32 +74,33 @@ class stream_transport : public stream_transport_base<Application> {
...
-        auto err = get<sec>(ret);
-        if (err == sec::unavailable_or_would_block) {
-          break;
-        } else {
-          CAF_LOG_DEBUG("read failed" << CAF_ARG(err));
-          this->next_layer_.handle_error(err);
-          return false;
-        }
+      } else if (num_bytes == 0) {
+        auto err = sec::socket_disconnected;
```
jakobod, line 95: Both cases run essentially identical code on error. Wouldn't a lambda be a cleaner solution?
```
auto handle_error = [this](sec err) -> bool {
  CAF_LOG_DEBUG("read failed" << CAF_ARG(err));
  this->next_layer_.handle_error(err);
  return false;
}

...

if (num_bytes == 0) {
  return handle_error(sec::socket_disconnected);
} else if (!last_socket_error_is_temporary()) {
  return handle_error(sec::socket_operation_failed);
}
```

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -131,20 +132,23 @@ class stream_transport : public stream_transport_base<Application> {
...
             recycle();
             written_ = 0;
           }
-        } else {
-          auto err = get<sec>(write_ret);
-          if (err != sec::unavailable_or_would_block) {
-            CAF_LOG_DEBUG("send failed" << CAF_ARG(err));
-            this->next_layer_.handle_error(err);
-          }
+        } else if (num_bytes == 0) {
```
jakobod, line 143: Same thing as above: I think a lambda could clean this up.

PR #87, opened 2020-10-01
Move message queue to Consumer

jakobod:
This PR moves the message and event queues from its previous location in the `endpoint_manager` to the consumer.

This PR is based on #86 and should be merged after that PR has been merged.


jakobod:
Sorry for the second rebase. I used an old `master` for the first one..
Ready for review now though!

libcaf_net/caf/net/consumer.hpp
```diff
@@ -19,17 +19,22 @@
...
 namespace caf::net {
 
-template <class Transport>
-endpoint_manager_ptr make_endpoint_manager(const multiplexer_ptr& mpx,
-                                           actor_system& sys, Transport trans) {
-  using impl = endpoint_manager_impl<Transport>;
-  return make_counted<impl>(mpx, sys, std::move(trans));
-}
+/// An implementation of BASP as an application layer protocol.
+class CAF_NET_EXPORT consumer {
```
Neverlord, line 27: What is this good for and why do we need it? It seems to me like the proxies (they hold a pointer to this interface) could simply have a pointer to `basp::application` instead. Also avoiding dynamic dispatching entirely.

jakobod: The idea was to not include `basp` in the proxy. In hindsight this approach is not optimal. I can remove the interface and exchange the pointer for `basp::application*`.

libcaf_net/caf/net/consumer_queue.hpp
```diff
@@ -34,7 +34,7 @@
 
 namespace caf::net {
 
-class CAF_NET_EXPORT endpoint_manager_queue {
+class CAF_NET_EXPORT consumer_queue {
```
Neverlord, line 37: The name suggests to me that this class is very generic queue type for (any number of?) producers to write to that belongs to a single consumer. But that just isn't the case. The implementation is in fact very specifically tailored to the needs of the BASP application. I think it shouldn't be in the public `caf::net` API to begin with and the name gives no guidance as to what role this class actually addresses.

libcaf_net/caf/net/stream_transport.hpp
```diff
@@ -155,6 +159,7 @@ class stream_transport {
       CAF_LOG_ERROR("send_buffer_size: " << socket_buf_size.error());
       return std::move(socket_buf_size.error());
     }
+    owner->register_reading();
```
Neverlord, line 162: Why should the transport unconditionally decide that the application *has* to read something right away?

Neverlord (changes requested):
I didn't look into the code in much detail yet.

May I suggest a change of perspective? The PR title reads like this is a simple refactoring step, while the proposed changes completely change the way BASP operates. Hence, this is in fact a major redesign. Shifting BASP to a message-oriented design opens up a lot of design space that seems mostly ignored. As a result, many pieces seem out of place to me and if there has been some thought put into it, there isn't any documentation or comment that communicates it.

On a side note, we moved away from the endpoint-manager design, because it forced BASP-specific notations into the lower API layers. Reusing code may be a good thing, but only after establishing a cohesive design.

It may also be a good idea to file non-architectural changes separately. For example, it seems like `length_prefix_framing` received a lot of changes (fixes?) and there's a new `message_oriented_layer_ptr` class. Those changes ought to be much less impactful and discussing/merging these changes may help keeping this PR focused.

libcaf_net/caf/net/basp/application.hpp
```diff
@@ -75,116 +84,436 @@ class CAF_NET_EXPORT application {
...
+    if (mailbox_.empty())
+      mailbox_.try_block();
+    return result;
+  }
 
-  error handle(size_t& next_read_size, packet_writer& writer, byte_span bytes);
+  // -- handling of outgoing messages and events -------------------------------
+
+  template <class LowerLayerPtr>
+  error dequeue_events(LowerLayerPtr& down) {
```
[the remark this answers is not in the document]
jakobod: How about `consume_events`/`consume_messages` then? That would match the naming scheme used within actors and the new `actor_shell`.

libcaf_net/caf/net/stream_transport.hpp
jakobod: The lifetime of `socket_manager` instances is bound to them being registered for reading or writing. As soon as a `socket_manager` is not registered for either of both, it is dereferenced by the multiplexer and hence, removed. Managers are only registered for writing, when they actually have to write data which leaves registering any manager for reading to keep them from being removed right away.

libcaf_net/caf/net/actor_proxy_impl.hpp
```diff
@@ -37,7 +37,7 @@ class actor_proxy_impl : public actor_proxy {
   void kill_proxy(execution_unit* ctx, error rsn) override;
 
 private:
-  endpoint_manager_ptr dst_;
+  basp::application* app_;
```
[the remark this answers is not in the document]
jakobod: > How is CAF making sure that the proxy never accesses a pointer past the lifetime of the pointed-to object now

It doesn't. The proxy would have to hold a `socket_manager_ptr` to ensure the lifetime of the BASP app, which is definitely possible. I'll add that.

libcaf_net/caf/net/basp/application.hpp
```diff
@@ -75,116 +84,436 @@ class CAF_NET_EXPORT application {
...
-  error handle_data(Parent& parent, byte_span bytes) {
-    static_assert(std::is_base_of<packet_writer, Parent>::value,
-                  "parent must implement packet_writer");
-    size_t next_read_size = header_size;
-    if (auto err = handle(next_read_size, parent, bytes))
-      return err;
-    parent.transport().configure_read(receive_policy::exactly(next_read_size));
-    return none;
+  template <class LowerLayerPtr>
+  bool prepare_send(LowerLayerPtr& down) {
```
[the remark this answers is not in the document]
jakobod: There is no benchmark that gives such distinct insight into the stack. At least not as of right now.
I used this approach, since it was used `libcaf_io` with the `basp_broker`.

PR #88, opened 2020-11-04
Implement statically typed actor_shell version

Neverlord:
Unlike with the old broker API, adding static type checking to the new actor shell API is very straightforward. 🙂


jakobod (approved):
LGTM apart from one minor comment.

PR #90, opened 2021-01-29
Fix build type propagation

josephnoir:
The build type is not propagated to the `CMakeCache.txt file`. I used `grep -rn CMAKE_BUILD_TYPE build` to check. This should fix it.


jakobod (approved):
👍

